PR #6, opened 2021-02-19
Add link to download html zip if available

nrdg42:
*Issue #, if available:*

*Description of changes:*
* Add link to download html zip if available
* Move run.json link from footer to new downloads area

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -310,12 +310,12 @@ def render(run, report_dir):
 
     svgs = render_runtimes(run, env, report_dir)
 
-    litani_zip_path = os.environ.get("LITANI_ZIP_PATH")
+    litani_report_archive_path = os.getenv("LITANI_REPORT_ARCHIVE_PATH")
 
     dash_templ = env.get_template("dashboard.jinja.html")
     page = dash_templ.render(
         run=run, svgs=svgs, litani_hash=get_git_hash(),
-        litani_version=litani.VERSION, litani_zip_path=litani_zip_path)
+        litani_version=litani.VERSION, litani_report_archive_path=litani_report_archive_path)
     with litani.atomic_write(report_dir / "index.html") as handle:
         print(page, file=handle)
 
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -310,10 +310,12 @@ def render(run, report_dir):
 
     svgs = render_runtimes(run, env, report_dir)
 
+    litani_zip_path = os.environ.get("LITANI_ZIP_PATH")
+
     dash_templ = env.get_template("dashboard.jinja.html")
     page = dash_templ.render(
         run=run, svgs=svgs, litani_hash=get_git_hash(),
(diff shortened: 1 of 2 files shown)


templates/dashboard.jinja.html
```diff
@@ -361,6 +390,25 @@ <h1>
...
+          href="run.json">run.json</a></code>
+        </p>
+      </li>
+      {% if litani_zip_path is not none %}
+      <li>
+        <p>HTML report: <code><a
+          href={{litani_zip_path}} download>{{litani_zip_path}}</a></code>
+        </p>
+      </li>
+      {% endif %}
```
karkhaz: Could we follow the
```
{% endif %}{# litani_zip_path is not none #}
```
here, these blocks eventually get quite hard to follow

templates/dashboard.jinja.html
```diff
@@ -361,6 +390,25 @@ <h1>
...
+      <li>
+        <p>
+          Raw data for this run: <code><a
+          href="run.json">run.json</a></code>
+        </p>
+      </li>
+      {% if litani_zip_path is not none %}
+      <li>
+        <p>HTML report: <code><a
+          href={{litani_zip_path}} download>{{litani_zip_path}}</a></code>
```
karkhaz: Jinja style is to have whitespace within expression and statements: `{{ litani_zip_path }}` rather than `{{litani_zip_path}}`. I prefer this, it makes it standout from the text

karkhaz: `href={{litani_zip_path}} download>` looks incorrect, do you mean `href={{litani_zip_path}}>download`?

lib/litani_report.py
```diff
@@ -310,10 +310,12 @@ def render(run, report_dir):
 
     svgs = render_runtimes(run, env, report_dir)
 
+    litani_zip_path = os.environ.get("LITANI_ZIP_PATH")
```
karkhaz: This is a bit generic, it'll get confusing if we have other zip files in the future (which I expect we will). Can I suggest `LITANI_REPORT_ARCHIVE_PATH`?

karkhaz (changes requested):

lib/litani_report.py
karkhaz: Also `os.getenv` is shorter than `os.environ.get`

templates/dashboard.jinja.html
karkhaz: And actually to be good HTML it should really be `href="{{ litani_zip_path }}">download`

nrdg42: Download is actually an [anchor tag attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a). Let me know if that is not what we want here.

karkhaz: Oh, very cool! Thanks! But I still think there should be quotes around the href `href="{{ litani_zip_path }}"`

karkhaz (approved):

PR #7, opened 2021-02-22
Symlink html report directory to temporary directory for writing and update atomically on render

nrdg42:
*Issue #, if available:*

*Description of changes:*
Symlink html report directory to temporary directory for writing and update atomically on render

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -300,8 +300,12 @@ def get_git_hash():
 
 
 def render(run, report_dir):
-    temporary_report_dir = pathlib.Path(tempfile.gettempdir()) / str(uuid.uuid4())
-    render_artifact_indexes(temporary_report_dir / "artifacts")
+    temporary_report_dir = litani.get_report_data_dir() / str(uuid.uuid4())
+
+    artifact_dir = temporary_report_dir / "artifacts"
+    shutil.copytree(litani.get_artifacts_dir(), artifact_dir)
+
+    render_artifact_indexes(artifact_dir)
 
     template_dir = pathlib.Path(__file__).parent.parent / "templates"
     env = jinja2.Environment(loader=jinja2.FileSystemLoader(str(template_dir)))
@@ -314,7 +318,7 @@ def render(run, report_dir):
     page = dash_templ.render(
         run=run, svgs=svgs, litani_hash=get_git_hash(),
         litani_version=litani.VERSION, litani_report_archive_path=litani_report_archive_path)
-    with litani.atomic_write(report_dir / "index.html") as handle:
+    with litani.atomic_write(temporary_report_dir / "index.html") as handle:
         print(page, file=handle)
 
     with litani.atomic_write(temporary_report_dir / litani.RUN_FILE) as handle:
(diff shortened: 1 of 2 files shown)


lib/litani_report.py
```diff
@@ -299,16 +300,13 @@ def get_git_hash():
 
 
 def render(run, report_dir):
-    artifacts_dst = report_dir / "artifacts"
-    if artifacts_dst.exists():
-        shutil.rmtree(report_dir / "artifacts")
-    shutil.copytree(litani.get_artifacts_dir(), report_dir / "artifacts")
-    render_artifact_indexes(artifacts_dst)
+    temporary_report_dir = pathlib.Path(tempfile.gettempdir()) / str(uuid.uuid4())
```
karkhaz: This is going to put the HTML directories under a temporary directory, where we won't later be able to find them. This will be a problem when we want to garbage-collect expired HTML directories.

I suggest you create an entirely new directory under the cache directory, perhaps `report_data`. So you'd have a method in `lib/litani.py` called `get_report_data_dir`. And you'd write these HTML directories to `get_report_data_dir() / str(uuid.uuid4())`.

That way, we can later simply iterate over all the HTML directories in there and delete them one by one (if they aren't locked). The cache directory is unique to a single running instance of litani, so it also means that several runs of litani won't garbage-collect each other's HTML directories.

So in summary: the cache directory would have a `html` and `report_data`. `html` at all times would be a symbolic link to one of the *subdirectories* of `report_data`.

lib/litani_report.py
```diff
@@ -319,22 +317,30 @@ def render(run, report_dir):
     with litani.atomic_write(report_dir / "index.html") as handle:
```
karkhaz: `temporary_report_dir`?

lib/litani_report.py
```diff
@@ -319,22 +317,30 @@ def render(run, report_dir):
...
         print(json.dumps(run, indent=2), file=handle)
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
         page = pipe_templ.render(run=run, pipe=pipe)
-        with litani.atomic_write(report_dir / pipe["url"]) as handle:
+        with litani.atomic_write(temporary_report_dir / pipe["url"]) as handle:
             print(page, file=handle)
 
+    temp_symlink_dir = report_dir.with_name(report_dir.name + "-tmp")
```
karkhaz: For total safety, I recommend to not use a fixed `-tmp` suffix. If this method somehow gets called twice at the same time, one of the `rename` calls will fail. Instead, use a UUID as a suffix. Since it's a temporary directory that you're going to rename in just a moment, it doesn't really matter. And if this gets called twice in parallel, one of the calls will 'win' without any failure.

lib/litani_report.py
```diff
@@ -319,22 +317,30 @@ def render(run, report_dir):
...
+    os.rename(temp_symlink_dir, report_dir)
+
 
 def render_artifact_indexes(artifact_dir):
     def dirs_needing_indexes():
         for root, dirs, fyles in os.walk(artifact_dir):
             if "index.html" not in fyles:
                 yield pathlib.Path(root), dirs, fyles
 
+    if artifact_dir.exists():
```
karkhaz: I'm not sure that this check is necessary? `artifact_dir` is now underneath a temporary directory that you've just created. This condition seems impossible to me.

lib/litani_report.py
```diff
@@ -319,22 +317,30 @@ def render(run, report_dir):
...
+        shutil.rmtree(artifact_dir)
+    shutil.copytree(litani.get_artifacts_dir(), artifact_dir)
```
karkhaz: Is there a good reason to move this line to this method? It used to be in `render`. I feel like copying the artifacts directory over is a different thing from rendering the indexes. To me it makes more sense that the client should first copy the directory, then render the indexes.

karkhaz (changes requested):

karkhaz (approved):

PR #8, opened 2021-02-24
Add print-capabilities top level flag

nrdg42:
*Issue #, if available:*

*Description of changes:*
Add print-capabilities top level flag that prints out a list of features in either human or machine-readable format. The intention is that any new API-changing features will be accompanied by a new entry in this list.


litani
```diff
@@ -191,6 +192,16 @@ def get_args():
         flags = arg.pop("flags")
         graph_pars.add_argument(*flags, **arg)
 
+    caps_pars = subs.add_parser("print-capabilities",
+        help="Print out Litani's capabilities in a list")
```
karkhaz, line 196: Lower case `p` in `Print` please 😛

eidelmanjonathan (approved):

PR #9, opened 2021-03-02
Garbage-collect old HTML directories

karkhaz:
Litani now marks the previous report directory as 'expired' immediately after writing a new one and moving the `html` symbolic link from the old to the new directory. Litani then deletes all expired report directories.

Since there is a risk that external processes might still be reading old report directories after having dereferenced the old `html` symlink, Litani now exposes an API for locking directories in a POSIX-compliant manner. External processes are expected to take a lock on report directories before attempting to access them. Litani will not delete expired directories as long as they are locked by external processes.

### DIFF
--- a/lib/capabilities.py
+++ b/lib/capabilities.py
@@ -16,6 +16,7 @@ import json
 
 _CAPABILITIES = {
     "atomic_report_update": "Report directory is rendered atomically",
+    "report_expire": "Old report directories will contain a .litani-expired file",
 }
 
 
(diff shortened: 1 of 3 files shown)


lib/report_directories.py
```diff
@@ -124,3 +124,19 @@ async def try_acquire_wait(self, timeout=0):
...
+def unlink_expired():
+    """Delete all report directories that are expired and unlocked"""
+
+    for data_dir in lib.litani.get_report_data_dir().iterdir():
+        lock_dir = LockableDirectory(data_dir)
+        if not lock_dir.acquire():
+            continue
+        expire_dir = ExpireableDirectory(data_dir)
+        if expire_dir.is_expired():
+            shutil.rmtree(lock_dir.path)
```
karkhaz: This would be clearer if it was just `shutil.rmtree(data_dir)`

nrdg42 (approved):

PR #11, opened 2021-03-08
Fix Atomic Lock API behavior

nrdg42:
*Issue #, if available:*

*Description of changes:*
Make it so that the LockableDirectory class is initially locked and has to call `release()`, just in case two processes are trying to Lock a Directory.

### DIFF
--- a/lib/litani.py
+++ b/lib/litani.py
@@ -65,11 +65,10 @@ class LockableDirectory:
 
 
     def __init__(self, path: pathlib.Path):
-        """Directory is initially released"""
+        """Directory is initially locked"""
 
         self.path = path.resolve()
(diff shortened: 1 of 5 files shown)


lib/litani.py
```diff
@@ -253,6 +252,7 @@ def unlink_expired():
             continue
         expire_dir = ExpireableDirectory(data_dir)
         if expire_dir.is_expired():
+            logging.debug("Unlinking %s", str(expire_dir))
```
karkhaz: This should be `data_dir`

karkhaz (changes requested):

karkhaz (approved):

PR #13, opened 2021-03-31
Add outcome table support

karkhaz:
This commit adds the --outcome-table flag to the `litani add-job`
command. This flag allows users to pass in an "outcome table" as a JSON
file, which litani uses to decide whether a job succeeded or failed.

### DIFF
--- a/litani
+++ b/litani
@@ -33,6 +33,7 @@ import uuid
 from lib import litani, litani_report, ninja_syntax
 import lib.capabilities
 import lib.graph
+import lib.job_outcome
 
 
 VALIDATE_DATA = False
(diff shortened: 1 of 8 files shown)


unit/outcome_table_decider.py
```diff
@@ -0,0 +1,251 @@
...
+    def test_one(self):
+        proc = unittest.mock.Mock()
+        proc.returncode = 1
+        jod = lib.job_outcome.OutcomeTableDecider(self.table, proc, False, True)
+
+        self.assertEqual(
+            jod.get_job_fields(), {
+                "outcome": "fail",
+                "loaded_outcome_dict": self.table,
+            })
```
adpaco: I think we could benefit from having an `assertOutcomeEquals(jod, outcome)` method that makes this call instead. You can then reuse it in many places throughout this code.

litani
```diff
@@ -687,26 +696,15 @@ def exec_job(args):
...
+        timeout_reached = True
 
-    out_data["command_return_code"] = proc.returncode
-    out_data["complete"] = True
-    ignore_returns = args.ignore_returns if args.ignore_returns else []
-    ignore_returns = [int(i) for i in ignore_returns]
-    ignore_returns.append(0)
-    if proc.returncode not in ignore_returns:
-        out_data["wrapper_return_code"] = 1
+    lib.job_outcome.fill_in_result(proc, timeout_reached, out_data, args)
```
adpaco, line 707: Nice!

PR #14, opened 2021-04-27
Add --output-directory and --output-symlink flags

nrdg42:
Add --output-directory and --output-symlink flags to litani init

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/litani
+++ b/litani
@@ -184,6 +184,14 @@ def get_args():
             "required": True,
             "help": "Project that this proof run is associated with",
             "metavar": "NAME",
+        }, {
+            "flags": ["--output-directory"],
+            "help": "Write Litani output to specified directory",
+            "metavar": "DIR"
+        }, {
+            "flags": ["--output-symlink"],
+            "help": "Create a symbolic link from DIR to litani's output directory",
+            "metavar": "DIR"
     }]:
         flags = arg.pop("flags")
         run_build_pars.add_argument(*flags, **arg)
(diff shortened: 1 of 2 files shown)


litani
```diff
@@ -537,13 +545,23 @@ def init(args):
...
         run_id = str(uuid.uuid4())
 
-    cache_dir = pathlib.Path(
-        tempfile.gettempdir()) / "litani" / "runs" / run_id
+    if args.output_directory:
+        output_directory = pathlib.Path(args.output_directory).resolve()
+    else:
+        output_directory = pathlib.Path(tempfile.gettempdir())
+
+    cache_dir = output_directory / "litani" / "runs" / run_id
```
karkhaz: I actually had in mind that `cache_dir` would be the path that the user provided, not a subdirectory 3 levels beneath it.

However, there is an advantage to doing it your way: users would not have to specify a fresh directory every time. I was assuming that users would do something like
```
litani init --output-directory output/$(date +%Y-%m-%dT%H:%M:%S)
```
and then the files would be written directly to that directory. Both are valid approaches, I suppose.

Here's what I suggest.
* Make `--output-directory` do what I'm suggesting.
* If you think it's worth it, add a third flag `--output-prefix` which does what you're doing up there. I do think it's nice to have, although I'm slightly wary of having too many flags, and I think what I proposed does give users more control.
* The two flags should be mutually exclusive.

karkhaz (changes requested):
Thanks for this! Please could you also add a new "capability" that indicates that these flags are available?

litani
```diff
@@ -537,13 +554,32 @@ def init(args):
...
 
-    cache_dir = pathlib.Path(
-        tempfile.gettempdir()) / "litani" / "runs" / run_id
-    cache_dir.mkdir(parents=True)
+    if args.output_prefix:
+        output_prefix = pathlib.Path(args.output_prefix).resolve()
+    else:
+        output_prefix = pathlib.Path(tempfile.gettempdir())
+
+    if args.output_directory:
```
karkhaz, line 557: Since `--output-prefix` and `--output-directory` are mutually exclusive, it would be good to reflect that here. Especially since if `--output-directory` is set, you're still assigning the `output_prefix` variable on line 560 and never using it, it just makes the code less obvious to read.

```python
if args.output_directory:
    cache_dir = pathlib.Path(args.output_directory).resolve()
else:
  if args.output_prefix:
      output_prefix = pathlib.Path(args.output_prefix).resolve()
  else:
      output_prefix = pathlib.Path(tempfile.gettempdir())
  cache_dir = output_prefix / "litani" / "runs" / run_id
```

litani
```diff
@@ -537,13 +554,32 @@ def init(args):
...
+        cache_dir = pathlib.Path(args.output_directory).resolve()
+    else:
+        cache_dir = output_prefix / "litani" / "runs" / run_id
+
+    try:
+        cache_dir.mkdir(parents=True)
+    except FileExistsError as err:
+        logging.error("Output directory already exists")
+        logging.error(str(err))
```
karkhaz: I don't particularly like printing stringified exceptions for this sort of user-facing tool, I'm not sure that it adds much value and it's more output for users to read/get confused by. I'd prefer just `logging.error("Output directory '%s' already exists", cache_dir)`.

karkhaz (changes requested):

karkhaz (approved):
Thank you!

PR #28, opened 2021-07-23
Add litani dump-run command

karkhaz:
This is to give litani runs the ability to 'introspect', that is, to
read the run.json file while the run is still in progress. This means
that clients do not have to know how to locate the cache directory, and
also means that we can guarantee that clients will see an up-to-date
version of the run. That is, if a Litani job runs litani dump-run, the
JSON file that gets printed will show all of the job's
reverse-dependencies has having completed. This is exercised in the e2e
test for this feature.

### DIFF
--- a/doc/src/man/litani-dump-run.scdoc
+++ b/doc/src/man/litani-dump-run.scdoc
@@ -1,0 +1,62 @@
+litani-dump-run(1) "" "Litani Build System"
+
+; Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+; Licensed under the Apache License, Version 2.0 (the "License").
+; You may not use this file except in compliance with the License.
+; A copy of the License is located at
+
+;     http://www.apache.org/licenses/LICENSE-2.0
+
+; or in the "license" file accompanying this file. This file is distributed
+; on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+; express or implied. See the License for the specific language governing
+; permissions and limitations under the License.
+
+
+# NAME
+
+litani dump-run - Print the current run as a JSON document to stdout
+
+
+# SYNOPSIS
+
+litani dump-run [OPTIONS]
+
+
+# DESCRIPTION
+
+This program prints the _run.json_ file for an in-progress Litani run to stdout.
+The JSON file's schema is documented in *litani-run.json(5)*.
+
+This program is intended to be used while an invocation of *litani-run-build(1)*
+is running. After the *run-build* process has terminated, this program will not
+work (though this may change in future versions). Instead, you can access the
+_run.json_ file in the run's output directory.
+
+This program may be run as a Litani job, and it may be run as a subprocess from
+a command that is part of a Litani job. This allows commands to 'introspect' on
+a Litani run so far. It is guaranteed that if a Litani job runs *litani
+dump-run*, all of the reverse-dependencies of that job will have the
+_"complete"_ key set to *True* in the printed _run.json_ file. That is, any job
+that runs *litani dump-run* will always see the most up-to-date state of the
+build with respect to its own reverse dependencies.
+
+
+# OPTIONS
+
+*-r* _N_, *--retries* _N_
+	How many times to retry loading the run file if the *litani run-build* process
+	doesn't emit it immediately. *litani dump-run* will retry loading the run once
+	every second up to the limit given in this flag. If _N_ is *0*, this program
+	will retry indefinitely---this is not recommended.
+
+
+# OUTPUT
+
+A JSON document printed to stdout. This will be either *null* (if this program was not able to load the run after a certain number of tries), or a document whose schema matches the one in *litani-run.json(5)*.
+
+
+# RETURN CODE
+
+Zero unless the program crashed.
--- a/lib/run_printer.py
+++ b/lib/run_printer.py
@@ -1,0 +1,87 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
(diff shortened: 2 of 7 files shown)


litani
```diff
@@ -371,7 +388,7 @@ def positive_int(arg):
     except ValueError as e:
         raise argparse.ArgumentTypeError(
             "Timeout '%s' must be an int" % arg) from e
-    if ret <= 0:
+    if ret < 0:
         raise argparse.ArgumentTypeError(
             "Timeout '%d' must be > 0" % ret) from e
```
nrdg42: Should you update the error to say must be `>= 0`?

test/e2e/tests/dump_run.py
```diff
@@ -0,0 +1,120 @@
...
+def get_init_args():
+    return {
+        "kwargs": {
+            "project": "foo",
+        }
+    }
+
+
+def get_jobs():
+    litani_path = pathlib.Path(__file__).parent.parent.parent.parent / "litani"
```
nrdg42: the `.parent.parent.parent.parent` is a bit of a red flag for me here. Is there a less hacky way of doing this?

lib/run_printer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def _exit_success(run):
+    print(json.dumps(run, indent=2))
+    sys.exit(0)
+
+
+def _exit_error():
+    print(json.dumps(None))
+    sys.exit(0)
```
angelonakos: Should this be `sys.exit(1)` ?

lib/run_printer.py
```diff
@@ -0,0 +1,87 @@
...
+    except FileNotFoundError:
+        _exit_error()
+
+    os.kill(pid, DUMP_SIGNAL)
+
+    cache_dir = lib.litani.get_cache_dir()
+
+    if args.retries:
+        for _ in range(args.retries):
+            _try_dump_run(cache_dir)
```
angelonakos: Should retries be spaced out by an interval?

lib/run_printer.py
karkhaz: Litani doesn't ever exit with non-zero unless it crashed (or unless the user asks it to, `litani run-build` has a flag to do that but it's not the default). This is because we usually have an in-band way of indicating failure. In this case, printing `null` is sufficient to indicate an error, because `null` is not a valid format for run.json. This is also mentioned in the man page.

litani
karkhaz: Done

test/e2e/tests/dump_run.py
karkhaz: Yes :) I fixed this so that the top-level end-to-end test runner sets the path to Litani as an environment variable, since it already has that information.

lib/run_printer.py
karkhaz: They are spaced out, the delay is in the `_try_dump_run` method. I put it there because there are two different calls to that function, so it made sense to put it in one place.

doc/src/man/litani-dump-run.scdoc
```diff
@@ -0,0 +1,62 @@
...
+*-r* _N_, *--retries* _N_
+	How many times to retry loading the run file if the *litani run-build* process
+	doesn't emit it immediately. *litani dump-run* will retry loading the run once
+	every second up to the limit given in this flag. If _N_ is *0*, this program
+	will retry indefinitely---this is not recommended.
+
+
+# OUTPUT
+
+A JSON document printed to stdout. This will be either *null* (if this program was not able to load the run after a certain number of tries), or a document whose schema matches the one in *litani-run.json(5)*.
```
angelonakos: I can't find a `litani-run.json` within kk-print-run (https://github.com/karkhaz/aws-build-accumulator/find/kk-print-run). Where is this JSON file located?

karkhaz: Thanks, this refers to the run.json man page, which I've posted a PR for here: https://github.com/awslabs/aws-build-accumulator/pull/32

lib/run_printer.py
```diff
@@ -0,0 +1,87 @@
...
+        os.unlink(cache_dir / _DUMPED_RUN)
+        _exit_success(run)
+    except (FileNotFoundError, json.decoder.JSONDecodeError):
+        time.sleep(1)
+
+
+async def dump_run(args):
+    try:
+        pid = lib.pid_file.read()
+    except FileNotFoundError:
```
angelonakos: Can an exception other than `FileNotFoundError` be thrown? What about capturing `OSError` instead? 

https://docs.python.org/3/library/exceptions.html#OSError

litani
```diff
@@ -371,9 +388,9 @@ def positive_int(arg):
     except ValueError as e:
         raise argparse.ArgumentTypeError(
             "Timeout '%s' must be an int" % arg) from e
-    if ret <= 0:
+    if ret < 0:
```
angelonakos: Technically, we could rename this to `non_negative_int`.

litani
```diff
@@ -236,6 +239,19 @@ def get_args():
...
 
+    dump_run_pars = subs.add_parser("dump-run")
+    dump_run_pars.set_defaults(func=lib.run_printer.dump_run)
+    for arg in [{
+            "flags": ["-r", "--retries"],
+            "metavar": "N",
+            "type": positive_int,
+            "default": 10,
+            "help":
+                "how many times to retry loading the run (Default: %(default)s)"
```
angelonakos: What about:

`"how many times to retry loading the run in intervals of one second (Default: %(default)s)"`

just in case somebody like myself missed that bit from the `man` page?

litani
karkhaz: Done in 0149ddd

litani
karkhaz: Done in 4ba87e5

lib/run_printer.py
angelonakos: [This](https://github.com/awslabs/aws-build-accumulator/blob/release/lib/litani.py#L182-L186) answers my own question.

angelonakos (approved):

PR #30, opened 2021-08-01
Add --phony-outputs flag

### LINKED ISSUE #29: Add --phony-output flag to litani-add-job
This flag should suppress the warning that Litani prints out when it doesn't find one of the output files upon completion of a job.

karkhaz:
This commit introduces a new flag to indicate that one or more of a
job's output files are 'phony', meaning that the job will not actually
write the files upon completion.

By default, Litani warns the user when a job has not written one of its
output files, because this can indicate that the build is incorrect. If
a job does not write its output files, then all dependent jobs will run
unconditionally, making the build unnecessarily slow. However, sometimes
users use outputs merely as a way to create a dependency between two
jobs, even when the first job doesn't actually write the output file. In
this case, the --phony-outputs flag can be used to suppress the warning,
so that only genuinely missing output files generate a warning.

This flag means that users no longer need to write 'touchfiles' upon job
completion to generate a dependency ordering. Instead, they can create
build graphs whose input and output nodes are entirely phony, enabling
the creation of dependency chains without having to manage temporary
files.

This commit fixes #29.

### DIFF
--- a/litani
+++ b/litani
@@ -34,6 +34,7 @@ from lib import litani, litani_report, ninja_syntax
 import lib.capabilities
 import lib.graph
 import lib.job_outcome
+import lib.output_artifact
 import lib.ninja
 import lib.process
 import lib.validation
@@ -144,6 +145,11 @@ def get_add_job_args():
             "default": 10,
             "type": int,
             "help": "seconds between memory profile polls"
+        }, {
+            "flags": ["--phony-outputs"],
+            "metavar": "OUT",
+            "nargs": "*",
+            "help": "do not warn if OUT does not exist upon job completion"
         }, {
             "flags": ["--tags"],
             "metavar": "TAG",
@@ -409,7 +415,7 @@ def make_litani_exec_command(add_args):
     # lists
     for arg in [
             "inputs", "outputs", "ignore_returns", "ok_returns",
-            "tags",
+            "tags", "phony_outputs",
     ]:
         if arg not in add_args or add_args[arg] is None:
             continue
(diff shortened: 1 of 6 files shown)


litani
```diff
@@ -144,6 +145,11 @@ def get_add_job_args():
             "default": 10,
             "type": int,
             "help": "seconds between memory profile polls"
+        }, {
+            "flags": ["--phony-outputs"],
+            "metavar": "OUT",
+            "nargs": "*",
```
angelonakos, line 151: Should this be changes to `+` with the added warning to use a `touchfiles` directory for dependency ordering if one chooses to not use the `--phony-outputs` parameter?

test/unit/output_artifact.py
```diff
@@ -0,0 +1,71 @@
...
+        with self.assertRaises(lib.output_artifact.MissingOutput):
+            copier.copy_output_file("foo")
+
+
+    def test_all_phony(self):
+        job_args = {
+            "phony_outputs": [],
+        }
+        copier = lib.output_artifact.Copier(self.artifacts_dir, job_args)
+        copier.copy_output_file("foo")
```
angelonakos, line 53: Should a ```os.path.exists('foo')``` be added?

test/unit/output_artifact.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+    def test_different_phony(self):
+        job_args = {
+            "phony_outputs": ["bar"],
+        }
+        copier = lib.output_artifact.Copier(self.artifacts_dir, job_args)
+
+        with self.assertRaises(lib.output_artifact.MissingOutput):
+            copier.copy_output_file("foo")
```
angelonakos, line 63: Should a ```os.path.exists('foo')``` be added?

test/unit/output_artifact.py
karkhaz: Not sure what you mean? Nothing is actually being copied here, in the `setUp` function the underlying call to `shutil.copyfile` is being mocked out. This is just testing that the copy won't even be called if the user passes in `--phony-outputs`; if that function did get called, the mock would just raise.

angelonakos: Right, the artifacts directory is being mocked. No need to assert that a file was copied to a non-existent/mocked location.

angelonakos (approved):
LGTM!

PR #46, opened 2021-09-22
Copy cbmc-viewer report directory artifact to final report.

markrtuttle:
Litani was failing to copy the cbmc-viewer report directory to the artifact directory because it was calling shutil.copy on the directory instead of shutil.copytree.

### DIFF
--- a/lib/output_artifact.py
+++ b/lib/output_artifact.py
@@ -13,6 +13,8 @@
 
 
 import dataclasses
+import logging
+import os
 import pathlib
 import shutil
 
@@ -25,7 +27,7 @@ class MissingOutput(Exception):
 
 @dataclasses.dataclass
 class Copier:
-    """Copy output artifacts to a directory, raising if they don't exist"""
+    """Copy output artifacts to a directory, raising MissingOutput if they don't exist"""
 
     artifacts_dir: pathlib.Path
     job_args: dict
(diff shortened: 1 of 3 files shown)


lib/output_artifact.py
```diff
@@ -25,15 +27,22 @@ class MissingOutput(Exception):
...
 @dataclasses.dataclass
 class Copier:
-    """Copy output artifacts to a directory, raising if they don't exist"""
+    """Copy output artifacts to a directory, raising MissingOutput if they don't exist"""
 
     artifacts_dir: pathlib.Path
     job_args: dict
 
 
     def copy_output_file(self, fyle):
```
nrdg42: Should we rename this method `copy_output_artifact` now?

lib/output_artifact.py
```diff
@@ -25,15 +27,22 @@ class MissingOutput(Exception):
...
         try:
-            shutil.copy(fyle, self.artifacts_dir)
+            if os.path.isfile(fyle):
+                shutil.copy(fyle, self.artifacts_dir)
+                return
+            if os.path.isdir(fyle):
+                shutil.copytree(fyle, self.artifacts_dir, dirs_exist_ok=True)
+                return
+            logging.warning("Can't copy %s: expected a file or a directory", fyle)
```
nrdg42: I think this will cause a lot of logging if we are using phony outputs. Should this be moved to after line 47?

lib/output_artifact.py
markrtuttle: Done.

lib/output_artifact.py
markrtuttle: I just deleted the warning.  This silence is no worse than the original silence. The scenario motivating the warning involved no phony outputs, so weaving this into the exception handler would mess it up too much.

nrdg42 (approved):

PR #49, opened 2021-09-29
Add support for custom stage names

### LINKED ISSUE #20: Make CI stages configurable

adpaco-aws:
*Related issues:* Fixes #20 

*Description of changes:*

Adds a new flag `--stage-names` to `litani init` so that the user can provide custom names for stages.

Stage names are part of the cache now and default stage names are used if custom names are not given.

Unfortunately, I was not able to add an end-to-end tests because they are all part of the same run and their stage names differ. This is what I tried to include:

```
def get_init_args():
    return {
        "kwargs": {
            "project": "stage_names",
            "stage-names": "stage_1 stage_2 stage_3",
        }
    }

def get_jobs():
    return [{
        "kwargs": {
            "command": "sleep 2",
            "description": "first",
            "ci-stage": "stage_1",
            "pipeline": "siesta",
        }
    }, {
        "kwargs": {
            "command": "sleep 2",
            "description": "second",
            "ci-stage": "stage_2",
            "pipeline": "siesta",
        },
    }, {
        "kwargs": {
            "command": "sleep 2",
            "description": "third",
            "ci-stage": "stage_3",
            "pipeline": "siesta",
        },
    }]

def get_run_build_args():
    return {}

def check_run(run):
    jobs = run["pipelines"][0]["ci_stages"][0]["jobs"]
    
    first = [j for j in jobs if j["wrapper_arguments"]["description"] == "first"][0]
    second = [j for j in jobs if j["wrapper_arguments"]["description"] == "second"][0]
    third = [j for j in jobs if j["wrapper_arguments"]["description"] == "third"][0]

    return all((
        first["ci-stage"] == "stage_1",
        second["ci-stage"] == "stage_2",
        third["ci-stage"] == "stage_3",
    ))
```

I think this test could work (with some changes) if we could put it on a separate run.
The script version of this test should allow you to see a preview of these changes:
```
python3 litani init --project-name test --stage-names stage_1 stage_2 stage_3
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_1
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_2
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_3
python3 litani run-build
```

I would like to get this test in but I am not sure what is the best way. Any requests or suggestions are welcome!


angelonakos:
@adpaco-aws 

I'm seeing one issue. When I navigate to the HTML page for the "siesta" pipeline in the generated report for your provided example, I notice that the order of the custom stages being displayed doesn't match the order with which the jobs were added. Specifically, I saw that "stage_2" was followed by "stage_1", which was followed by "stage_3". This is because `--stage-names` accepts multiple arguments and doesn't maintain an order between the arguments.

I attempted to add an [`append` action](https://docs.python.org/3/library/argparse.html#action) to the `--stage-names` flag, however, that isn't enough as stages are still being rendered out of order.
Another direction could be changing the arguments being passed to `litani init`. We could add a `--first-stage` flag, which is marked as "required", and then add `--second-stage` and `--third-stage` flags. We will always render the "first stage" and then render "second stage" and "third stage" only if they were provided initially.

adpaco-aws:
Hi @angelonakos , as far as I know there is nothing wrong with the way stage names are passed as arguments. When I look into the `run.json` file, I can see stage names being stored in order:
```json
  "stages": [
    "stage_1",
    "stage_2",
    "stage_3"
  ]
```
Of course, we can change it so that stage names are passed individually if that is a preference on your end.

But I believe this particular issue is due to the way we sort jobs within a single pipeline.

adpaco-aws:
Sorry, it turns out this has nothing to do with job sorting.

The actual problem was that I changed this (in line 444 from `litani_report.py`):
```python
        for stage in litani.CI_STAGES:
```
to
```python
       for stage in pipe["ci_stages"]:
```
That was causing this particular issue.

But since we have in-order stages stored in the run, I have replaced it with:
```python
        for stage in run["stages"]:
```
Now it seems to be working as expected. Thanks for the feedback!

angelonakos:
@adpaco-aws  the latest changes did, in fact, yield the correct and expected order of stages for me.

litani
```diff
@@ -221,6 +219,12 @@ def get_args():
             "flags": ["--output-symlink"],
             "help": "create a symbolic link from DIR to litani's output directory",
             "metavar": "DIR"
+        }, {
+            "flags": ["--stage-names"],
+            "help":
+                "stage names for the project",
```
karkhaz:
```
"default": litani.DEFAULT_STAGES,
"help": "stages that a job can be a member of. Default: %(default)s",
```

litani
```diff
@@ -541,6 +546,19 @@ def get_pools(args):
 
     return ret
 
+def get_stages(args):
+    ret = litani.DEFAULT_STAGES
```
karkhaz: If you make this the default in the argument parser, you can delete these 3 lines

litani
```diff
@@ -650,7 +669,8 @@ async def run_build(args):
     rules = []
     builds = []
     pools = {}
-    fill_out_ninja(cache, rules, builds, pools)
+    stages = []
```
karkhaz: `stages` appears to be unused. You're passing it into `fill_out_ninja` which adds items to it, but nothing happens to it afterwards... I'm pretty sure that this as well as the changes to the `fill_out_ninja` function are not necessary

doc/src/man/litani-init.scdoc
```diff
@@ -54,6 +54,12 @@ of *litani add-job* will add jobs to a single run.
 	at the top of the HTML report. The project name is also included in the
 	_run.json_ file to help find all the runs for a particular project.
 
+*--stage-names _NAME_ _[NAME...]_
+	The stage names that this run is composed of. Stage names are printed in
+	above the first pipeline in the run. It is possible to provide up to three
+	names using this flag. Default stage names are used if less than three names
```
angelonakos: You could also mention the default stages of `build`, `test`, `report` here

karkhaz (changes requested):
I don't want to merge this as-is, as I think that this would force an API break in the future.

If we have a `--stage-names` flag, then it should be able to accept an arbitrary number of stages. If you pass 2 stages, then there should be only 2 stages on the dashboard. This implementation adds a third stage even if you pass two. The problem is that if we merge this PR, and then 'fix' it in the future, then the meaning of the `--stage-names` flag will change, which I'm not willing to do here.

To merge this, I think the flag should take a non-zero number of stage names. The dashboard should display only 1 or 2 columns if there are less than 3 stage names. If there are more than 3, I suggest something like a single long progress bar for the whole pipeline  rather than having too many columns.

Not sure if you have the bandwidth to implement all this...

adpaco-aws:
> I don't want to merge this as-is, as I think that this would force an API break in the future.
> 
> If we have a `--stage-names` flag, then it should be able to accept an arbitrary number of stages. If you pass 2 stages, then there should be only 2 stages on the dashboard. This implementation adds a third stage even if you pass two. The problem is that if we merge this PR, and then 'fix' it in the future, then the meaning of the `--stage-names` flag will change, which I'm not willing to do here.

No problem, this sounds reasonable. I went ahead and changed it so we accept a non-zero number of stages. It made more sense to call the flag `--stages`. Most things were working out of the box.

Finally figured out how to write a test, which passes locally for me, so added that too.

> If there are more than 3, I suggest something like a single long progress bar for the whole pipeline rather than having too many columns.

I do not think it looks that bad if you have 4-5 stages. My concern is that using a long bar will not help debugging. Are you sure about this?

Another concern: How do we validate that a job belongs to one of the stages now? I can add code for that at the beginning of `add_job`, but maybe there is a better way. What do you think?

@angelonakos I changed the documentation there, how does it look now?

angelonakos:
> @angelonakos I changed the documentation there, how does it look now?

LGTM

lib/validation.py
```diff
@@ -324,7 +326,7 @@ def _run_schema():
                 "complete": bool,
                 # Whether all the jobs in this stage are complete.
 
-                "name": voluptuous.Any("build", "test", "report"),
+                "name": str,
                 # The stage's name. In the future, it may be possible for users
```
karkhaz: So these comments actually get emitted into the man page :) Please could you update them, in particular "In the future, it may be possible for users to pass arbitrary names"

lib/validation.py
```diff
@@ -236,6 +236,8 @@ def _run_schema():
         # the HTML report generator and can be used to group related sets of
         # runs, but is otherwise not used by litani.
 
+        "stages": [str],
```
karkhaz, line 239: Please add a comment under this line. The comment will be rendered into the `litani-run.json` man page. Something like `"The CI stages that each job can be a member of"`.

doc/src/man/litani-init.scdoc
```diff
@@ -54,6 +54,11 @@ of *litani add-job* will add jobs to a single run.
 	at the top of the HTML report. The project name is also included in the
 	_run.json_ file to help find all the runs for a particular project.
 
+*--stages _NAME_ _[NAME...]_
+	The stages that this run is composed of. At least one stage has to be
+	provided. If this flag is not specified, the default stages `build`, `test`
```
karkhaz: Please use `_` rather than `, as this is scdoc rather than markdown.

test/e2e/tests/custom_stages.py
```diff
@@ -0,0 +1,57 @@
...
+                "pipeline": "siesta",
+            }
+        }
+        jobs.append(job)
+    return jobs
+
+def get_run_build_args():
+    return {}
+
+def check_run(run):
```
karkhaz, line 34: Nice, thanks a lot for this! Please ensure that there's a newline at the end of the file

karkhaz (changes requested):
This is really nice, thanks. Just a few minor comments. For reference, here's what this looks like with the 4 stages in your test suite:
Screenshot 2021-10-11 at 22 01 11

adpaco-aws:
Just pushed changes to address your comments. Let me know how it looks now.

Can you please also comment on this?
> Another concern: How do we validate that a job belongs to one of the stages now? I can add code for that at the beginning of add_job, but maybe there is a better way. What do you think?

angelonakos:
> Just pushed changes to address your comments. Let me know how it looks now.
> 
> Can you please also comment on this?
> 
> > Another concern: How do we validate that a job belongs to one of the stages now? I can add code for that at the beginning of add_job, but maybe there is a better way. What do you think?

Why would you need to validate that a job belongs to a particular stage? What use case do you have in mind? 

When the following is ran:

```bash
python3 litani init --project-name test --stage-names stage_1 stage_2 stage_3
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_1
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_2
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage stage_3
python3 litani run-build
```

3 different jobs are added to stage_1, stage_2 and stage_3 of the siesta pipeline respectively. 
Is there a need to confirm that the right command was passed to a particular stage and pipeline?

adpaco-aws:
Because it was being done before. Take the following script where there is a typo in one of `--ci-stage` commands:
```bash
python3 litani init --project-name original
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage build
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage teste
python3 litani add-job --command 'sleep 2' --pipeline-name 'siesta' --ci-stage report
python3 litani run-build
```
In the release version, this is what we get:
```
Report will be rendered at file:///var/folders/qb/tckm4wwd0gzbl9ypc53ncdb8xg9k_h/T/litani/runs/latest/html/index.html
usage: litani add-job [-h] [--inputs F [F ...]] --command C [--outputs F [F ...]] --pipeline-name P --ci-stage S [--cwd DIR] [--timeout N]
                      [--timeout-ok] [--timeout-ignore] [--ignore-returns RC [RC ...]] [--ok-returns RC [RC ...]] [--outcome-table F]
                      [--interleave-stdout-stderr] [--stdout-file FILE] [--stderr-file FILE] [--pool NAME] [--description DESC] [--profile-memory]
                      [--profile-memory-interval N] [--phony-outputs [OUT ...]] [--tags TAG [TAG ...]]
litani add-job: error: argument --ci-stage: invalid choice: 'teste' (choose from 'build', 'test', 'report')
[2/2] /Users/accorell/litani-stages/litani exec --command 'sleep 2' --pipeline-name siesta --ci-stage report --job-id b70f0338-cdb8-4ed9-a174-a8fa17...
Report was rendered at file:///var/folders/qb/tckm4wwd0gzbl9ypc53ncdb8xg9k_h/T/litani/runs/latest/html/index.html
```
The other jobs run as planned and the CI stage shows an empty bar because it has no jobs associated. However, we also get an error indicating the "invalid choice", which we do not generate anymore. I can picture myself spending half a day debugging an issue related to this, only to find that litani does not complain if the CI stage you pass is not valid.

angelonakos:
Got it. Adrian, that helped.

To ensure that a job is added to a stage that **actually** exists, can we not just add the following:

```python
    cache_file = litani.get_cache_dir() / litani.CACHE_FILE
    with open(cache_file) as handle:
        cache_contents = json.load(handle)
    if args.ci_stage not in cache_contents["stages"]:
        logging.error("Invalid stage name '%s' was provided", args.ci_stage)
        logging.error("Possible stage names: '%s'", "', '".join(cache_contents["stages"]))
        logging.error("Cannot add job: %s", json.dumps(job, indent=2))
        raise  argparse.ArgumentTypeError("Invalid arguments provided to litani add-job")
```

at L636 of litani in your branch?

Basically, inside of `async def add_job(args)` right after this line:

```python
    job["status_file"] = str(
        litani.get_status_dir() / ("%s.json" % job_id))
```

My test:
image

adpaco-aws:
Thanks for the quick response and suggestions!

Your proposal is similar to what I had in mind, which I just added to the PR. A few notes:
1. I believe validation should be the first thing to occur, which is why I added the code at the beginning of `add_job`.
2. I changed your suggestion to follow what seems to be the common way of handling errors in litani: One call to `logging_error` then another one to `sys.exit(1)`.

Please let me know if you see any issues with this.

angelonakos:
> I changed your suggestion to follow what seems to be the common way of handling errors in litani: One call to `logging_error` then another one to `sys.exit(1)`.

Nice!

angelonakos (approved):

karkhaz (approved):

karkhaz:
Merged. Thanks!

Issue #45, opened 2021-09-20
Un-truncate pipeline dependency graph file names


karkhaz:
File names in dep graphs should not be truncated. Instead, they should wrap to the next line after a certain number of characters. Ideally, the wrap split would be done on path separator (`/`) if possible.

ronakfof:
Fixed by PR [#56 ](https://github.com/awslabs/aws-build-accumulator/pull/56)

PR #53, opened 2021-10-07
Adding a line break after a specific truncate length

ronakfof:
*Issue #, if available:*
https://github.com/awslabs/aws-build-accumulator/issues/45

*Description of changes:*
Adding a line break after a specified truncate length

### DIFF
--- a/lib/graph.py
+++ b/lib/graph.py
@@ -34,6 +34,11 @@ class Node:
         return string
 
 
+def wrap_command(limit, command):
+    occurrences = re.findall('.{1,%i}' % limit, command)
+    newline_separated_command = '\n'.join(occurrences)
+    return newline_separated_command
+
 
 class DependencyNode(Node):
     def __init__(self, fyle, truncate=30, **style):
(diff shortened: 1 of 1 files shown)


lib/graph.py
```diff
@@ -74,7 +76,9 @@ def __init__(self, command, truncate=30, **style):
         self.style = style
 
         if len(command) > truncate:
-            self.style["label"] = command[:truncate-1] + "…"
+            self.style["label"] = '\n'.join(
+                re.findall('.{1,%i}' % truncate, command)
```
angelonakos: Why are you using `truncate` and not `truncate-1` in the modulo operation?

Meaning, why are you using:

```python
re.findall('.{1,%i}' % truncate, command)
```

instead of:


```python
re.findall('.{1,%i}' % (truncate-1), command)
```

angelonakos (changes requested):
I ran this locally and long labels were wrapped appropriately.

However, I suggest create a function such as the one below and reusing appropriately:

```python
def wrap_command(command, limit):
    occurrences = re.findall('.{1,%i}' % limit, command)
    newline_separated_command = '\n'.join(occurrences)
    return newline_separated_command
    
```

angelonakos (approved):

ronakfof:
Closing this one as the issue will be resolved with this [PR](https://github.com/awslabs/aws-build-accumulator/pull/56).

Issue #40, opened 2021-09-20
Make graph output optional


karkhaz:
Acceptance:
* `litani run-build` has a `--no-draw-dependency-graph` flag. When passed, the report generator does not attempt to generate dependency graphs.
* Measure difference in runtime on a proof project and include that in the commit message.

PR #54, opened 2021-10-07
Added --no-draw-dependency-graph flag to make graph output optional

ronakfof:
- Runtime with no graph generation took ~1-3 secs less than with graph generation

*Issue #, if available:*
https://github.com/awslabs/aws-build-accumulator/issues/40

*Description of changes:*
Added a **_--no-draw-dependency-graph_** flag to litani run-build to make graph output optional.
Noticed **~1-3secs** less runtime with no graph generation.

### DIFF
--- a/doc/src/man/litani-run-build.scdoc
+++ b/doc/src/man/litani-run-build.scdoc
@@ -25,6 +25,7 @@ litani run-build - Start a Litani run
 	\[*-j*/*--parallel* _N_]
 	\[*-o*/*--out-file* _F_]
 	\[*--fail-on-pipeline-failure*]
+	\[*--no-pipeline-dep-graph*]
 	\[*-p*/*--pipelines* _P_ [_P_ ...]]
 	\[*-s*/*--ci-stage* _S_]
 
@@ -59,6 +60,11 @@ parallel.
 *--fail-on-pipeline-failure*
 	Return _0_ only if all pipelines were successful. See *RETURN CODE*
 	below.
+
+*--no-pipeline-dep-graph*
+	Do not render dependency graphs for each pipeline onto the HTML individual
+	pipeline pages. Pipeline graphs will also not be rendered if Graphviz is
+	not installed.
 
 *-p* _P_ [_P_ ...], *--pipelines* _P_ [_P_ ...]
 	Only run jobs that are part of the specified pipelines.
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -75,13 +75,13 @@ class Gnuplot:
 
 
 class PipelineDepgraphRenderer:
-    def __init__(self):
-        self.should_render = shutil.which("dot") is not None
-
-        if not self.should_render:
+    def __init__(self, should_render=True):
+        is_graphviz_installed = shutil.which("dot") is not None
+        if not is_graphviz_installed:
             logging.info(
                 "Graphviz is not installed; pipeline dependency "
-                "graph will not be rendered")
+                "graph cannot be rendered")
+        self.should_render = should_render and is_graphviz_installed
 
 
     def render_to_file(self, out_file, pipe):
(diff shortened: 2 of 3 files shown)


litani
```diff
@@ -300,6 +300,10 @@ def get_args():
             "flags": ["--fail-on-pipeline-failure"],
             "action": "store_true",
             "help": "exit with a non-zero return code if some pipelines failed"
+    }, {
```
karkhaz, line 307: Please also add documentation to the manpage

lib/litani_report.py
```diff
@@ -547,9 +547,10 @@ def render(run, report_dir):
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
-            render_root=temporary_report_dir,
-            pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
+        if draw_dependency_graph:
```
karkhaz: I suggest to get #55 merged. Once you address [this comment](https://github.com/awslabs/aws-build-accumulator/pull/55#discussion_r726550269), you might find you can use the same mechanism here

litani
```diff
@@ -358,7 +362,7 @@ def get_args():
 
 
 
-def continuous_render_report(cache_dir, report_dir, killer, out_file):
+def continuous_render_report(cache_dir, report_dir, killer, out_file, draw_dependency_graph):
```
karkhaz: Line too long, break at `cache_dir`

litani
```diff
@@ -664,15 +668,15 @@ async def run_build(args):
         for build in builds:
             logging.debug(build)
             ninja.build(**build)
-
+    draw_dependency_graph = False if args.no_draw_dependency_graph else True
```
karkhaz: No need for this, see below

litani
```diff
@@ -664,15 +668,15 @@ async def run_build(args):
...
     report_dir = litani.get_report_dir()
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    litani_report.render(run, report_dir)
+    litani_report.render(run, report_dir, draw_dependency_graph)
```
karkhaz: `, not args.no_draw_dependency_graph`

karkhaz: Furthermore, if you finish #55 , you'll be able to pass this variable just once to a `PipelineDepgraphRenderer` and then pass the `PipelineDepgraphRenderer` object to all these calls

karkhaz (changes requested):

lib/litani_report.py
```diff
@@ -75,13 +75,15 @@ def render(self, gnu_file, out_file=None):
...
 
 class PipelineDepgraphRenderer:
-    def __init__(self):
-        self.should_render = shutil.which("dot") is not None
-
-        if not self.should_render:
-            logging.info(
-                "Graphviz is not installed; pipeline dependency "
-                "graph will not be rendered")
+    def __init__(self, should_render=True):
```
angelonakos, line 78: What do you think of using something such as the following instead?

```python
    def __init__(self, should_render=True):
        is_graphviz_installed = shutil.which("dot") is not None
        if not is_graphviz_installed:
            logging.info(
                "Graphviz is not installed; pipeline dependency "
                "graph cannot be rendered")
        self.should_render = should_render and is_graphviz_installed
```

angelonakos (approved):

angelonakos (approved):

doc/src/man/litani-run-build.scdoc
```diff
@@ -60,6 +61,9 @@ parallel.
 	Return _0_ only if all pipelines were successful. See *RETURN CODE*
 	below.
 
+*--no-draw-dependency-graph*
+	Does not attempt to generate dependency graph.
```
karkhaz: * I think this should be `--no-pipeline-dep-graph`, because there are other kinds of dependency graphs that we could generate in the future.
* Use the imperative: "Do not". I would write "Do not render dependency graphs for each pipeline onto the HTML individual pipeline pages. Pipeline graphs will also not be rendered if Graphviz is not installed".

karkhaz (changes requested):

karkhaz (approved):

Issue #41, opened 2021-09-20
Do not retry graph generation if `dot` is not installed


karkhaz:
Currently, the PipelineDepgraphRenderer class does not remember if the Graphviz `dot` command is installed. It repeatedly calls `dot` every time the report gets generated, catching the exception if `dot` is not installed. This is a waste of time, because forking is expensive, especially for hundreds of pipeline calls.

Acceptance:
* If `dot` is not installed, then the `render` method is a noop.
* Measure difference in runtime on a proof project and include that in the commit message.

PR #55, opened 2021-10-07
Added a check to render graph only if dot is installed

ronakfof:
*Issue #, if available:*
https://github.com/awslabs/aws-build-accumulator/issues/41

*Description of changes:*
Added a check to render graph only if dot is installed and noticed a decrease in runtime by ~1-3secs when dot is not installed.

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -74,31 +74,33 @@ class Gnuplot:
 # ______________________________________________________________________________
 
 
-@dataclasses.dataclass
 class PipelineDepgraphRenderer:
-    pipe: dict
+    def __init__(self):
+        self.should_render = shutil.which("dot") is not None
+
+        if not self.should_render:
+            logging.info(
+                "Graphviz is not installed; pipeline dependency "
+                "graph will not be rendered")
 
 
-    def render_to_file(self, out_file):
-        dot_graph = lib.graph.SinglePipelineGraph.render(self.pipe)
+    def render_to_file(self, out_file, pipe):
+        dot_graph = lib.graph.SinglePipelineGraph.render(pipe)
         out_file.parent.mkdir(exist_ok=True, parents=True)
         with open(out_file, "w") as handle:
-            try:
-                with subprocess.Popen(
-                        ["dot", "-Tsvg"], text=True, stdin=subprocess.PIPE,
-                        stdout=handle) as proc:
-                    proc.communicate(input=dot_graph)
-            except FileNotFoundError:
-                return False
+            with subprocess.Popen(
+                    ["dot", "-Tsvg"], text=True, stdin=subprocess.PIPE,
+                    stdout=handle) as proc:
+                proc.communicate(input=dot_graph)
         return not proc.returncode
 
 
-    @staticmethod
-    def render(render_root, pipe_url, pipe):
-        pdr = PipelineDepgraphRenderer(pipe)
+    def render(self, render_root, pipe_url, pipe):
+        if not self.should_render:
+            return
         out_rel = pipe_url / "dependencies.svg"
         out_file = render_root / out_rel
-        success = pdr.render_to_file(out_file=out_file)
+        success = self.render_to_file(out_file=out_file, pipe=pipe)
         if success:
             pipe["dependencies_url"] = "dependencies.svg"
 
@@ -511,7 +513,7 @@ def get_git_hash():
         return None
 
 
-def render(run, report_dir):
+def render(run, report_dir, pipeline_degraph_renderer):
     temporary_report_dir = litani.get_report_data_dir() / str(uuid.uuid4())
     temporary_report_dir.mkdir(parents=True)
     old_report_dir_path = litani.get_report_dir().resolve()
@@ -547,7 +549,7 @@ def render(run, report_dir):
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
+        pipeline_degraph_renderer.render(
             render_root=temporary_report_dir,
             pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
         for stage in pipe["ci_stages"]:
--- a/litani
+++ b/litani
@@ -360,7 +360,8 @@ def get_args():
 
 
 
-def continuous_render_report(cache_dir, report_dir, killer, out_file):
+def continuous_render_report(
+    cache_dir, report_dir, killer, out_file, pipeline_degraph_renderer):
     try:
         while True:
             run = litani_report.get_run_data(cache_dir)
@@ -370,7 +371,7 @@ def continuous_render_report(cache_dir, report_dir, killer, out_file):
             if out_file is not None:
                 with litani.atomic_write(out_file) as handle:
                     print(json.dumps(run, indent=2), file=handle)
-            litani_report.render(run, report_dir)
+            litani_report.render(run, report_dir, pipeline_degraph_renderer)
             if killer.is_set():
                 break
             time.sleep(2)
@@ -681,11 +682,13 @@ async def run_build(args):
     report_dir = litani.get_report_dir()
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    litani_report.render(run, report_dir)
+    pipeline_degraph_renderer = litani_report.PipelineDepgraphRenderer()
+    litani_report.render(run, report_dir, pipeline_degraph_renderer)
     killer = threading.Event()
     render_thread = threading.Thread(
         group=None, target=continuous_render_report,
-        args=(cache_dir, report_dir, killer, args.out_file))
+        args=(cache_dir, report_dir, killer, args.out_file,
+        pipeline_degraph_renderer))
     render_thread.start()
 
     runner = lib.ninja.Runner(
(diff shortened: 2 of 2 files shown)


lib/litani_report.py
```diff
@@ -547,9 +547,13 @@ def render(run, report_dir):
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
-            render_root=temporary_report_dir,
-            pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
+        if shutil.which("dot"):
```
karkhaz: This is still pretty inefficient. The problem is that `litani_report.render()` gets called repeatedly during the run, and within `litani_report.render()` this for loop iterates over every pipeline. This code is calling `which` on every single iteration, there's no need for that... if `dot` wasn't installed five seconds ago it's not going to be installed now either :)

I was hoping something more like: when you construct a `PipelineDepgraphRenderer`, it calls `which` just once to find if dot is installed. If it is not installed, then `PipelineDepgraphRenderer.render()` should become a no-op, so it doesn't matter that we call it repeatedly.

karkhaz (changes requested):

lib/litani_report.py
karkhaz: In fact, it might be better to construct a single `PipelineDepgraphRenderer` outside this method, i.e. in the actual `run_build` method at the top level. And then pass that object to this method. That way you really only have to construct it once.

lib/litani_report.py
```diff
@@ -547,9 +547,12 @@ def render(run, report_dir):
...
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
-            render_root=temporary_report_dir,
-            pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
+        if pipeline_degraph_renderer:
+            pipeline_degraph_renderer.render(
+                render_root=temporary_report_dir,
+                pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
+        else:
+            print("Please install dot if you wish to generate dependency graph")
```
karkhaz: We never use `print` for this kind of status message. Always `logging`. `print` is for output the user specifically asked for, for example `litani dump-run` prints out a JSON representation of the run using `print`. This is so that stuff that the user asks for goes to stdout, and status/debug/warning/error messages go to stderr. This way, users can redirect the output they want to a file by using `>`, but error messages still get printed to the terminal.

Also, putting the message here is also going to be print it out once every few seconds if the message is here, no need for that. Better to do it in the constructor for PipelineDepgraphRenderer as mentioned above.

lib/litani_report.py
```diff
@@ -547,9 +547,12 @@ def render(run, report_dir):
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
-            render_root=temporary_report_dir,
-            pipe_url=pathlib.Path(pipe["url"]), pipe=pipe)
+        if pipeline_degraph_renderer:
```
karkhaz: I still don't like doing this check here. It seems cleaner to me to pass in a PipelineDepgraphRenderer no matter what (so never pass None), but the `render` method of the object itself should just be a no-op if dot is not installed.

For example, the __init__ method of PipelineDepgraphRenderer could run `which`, and if dot isn't installed it would set an instance variable to False, and then the render method would check that variable and return early.

litani
```diff
@@ -668,11 +669,16 @@ async def run_build(args):
     report_dir = litani.get_report_dir()
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    litani_report.render(run, report_dir)
+    if shutil.which("dot"):
```
karkhaz: As above, please do this check in the constructor

karkhaz (changes requested):

angelonakos (approved):

lib/litani_report.py
```diff
@@ -94,7 +95,11 @@ def render_to_file(self, out_file):
 
 
     @staticmethod
-    def render(render_root, pipe_url, pipe):
+    def render(should_render, render_root, pipe_url, pipe):
+        if not should_render:
```
karkhaz: Why not just do `if not PipelineDepgraphRenderer.should_render`. Then you don't have to pass it in as an argument.

litani
```diff
@@ -668,11 +669,13 @@ async def run_build(args):
     report_dir = litani.get_report_dir()
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    litani_report.render(run, report_dir)
+    pipeline_degraph_renderer = litani_report.PipelineDepgraphRenderer
```
karkhaz: I don't understand this, why create a class object? Shouldn't you be creating an instance?

lib/litani_report.py
```diff
@@ -547,7 +552,7 @@ def render(run, report_dir):
 
     pipe_templ = env.get_template("pipeline.jinja.html")
     for pipe in run["pipelines"]:
-        PipelineDepgraphRenderer.render(
+        pipeline_degraph_renderer.render(pipeline_degraph_renderer.should_render,
```
karkhaz: As above: I don't think it's necessary to pass in an argument like this. It's also confusing to pass in a class as a variable, I think it was better how it was before.

karkhaz (changes requested):
There are still a bunch of problems here.

* I don't understand why we're passing a `should_render` value to the `render` method, the `PipelineDepgraphRenderer` class already has a `should_render` instance variable. It seems that `render` should no longer be a static method; `render` should check `self.should_render` instead of taking an argument.
* The `logging.info` is still going to get printed out every few seconds. I think the `__init__` method should do `which`, set the `should_render` instance variable, and print the message out once. You can then construct a PipelineDepgraphRenderer object once at the top level and pass it into `litani_report.render`.

lib/litani_report.py
```diff
@@ -94,7 +95,11 @@ def render_to_file(self, out_file):
 
 
     @staticmethod
-    def render(render_root, pipe_url, pipe):
+    def render(should_render, render_root, pipe_url, pipe):
+        if not should_render:
+            logging.info("Please install dot if you wish to generate \
```
karkhaz: No need for a line continuation. I would do
```
            logging.info(
                "Graphviz is not installed; pipeline dependency "
                "graph will not be rendered")
```

lib/litani_report.py
ronakfof: I did that earlier but i changed it to this as i made changes to this PR to support PR #54 as well. in that one i check for flag and update the ```should_render``` variable. I can roll this one back to how it was and we can discuss the changes needed on that one once this gets merged.

lib/litani_report.py
```diff
@@ -74,9 +74,15 @@ def render(self, gnu_file, out_file=None):
 # ______________________________________________________________________________
 
 
-@dataclasses.dataclass
 class PipelineDepgraphRenderer:
-    pipe: dict
+    def __init__(self, pipe=None):
```
angelonakos: Why was the default for `pipe` set to `None`?

ronakfof: As i am not passing any variable when i create an instance on top-level here ```pipeline_degraph_renderer = litani_report.PipelineDepgraphRenderer()``` on line 672 of ```litani``` file. Any suggestions?

angelonakos (approved):

lib/litani_report.py
```diff
@@ -74,9 +74,15 @@ def render(self, gnu_file, out_file=None):
 # ______________________________________________________________________________
 
 
-@dataclasses.dataclass
 class PipelineDepgraphRenderer:
-    pipe: dict
+    def __init__(self, pipe):
+        self.pipe = pipe
```
karkhaz: `self.pipe` is no longer being used. It should be removed. There's no need to pass anything in the constructor any more, since there's already a pipe passed into `render`.

lib/litani_report.py
```diff
@@ -93,8 +99,9 @@ def render_to_file(self, out_file):
         return not proc.returncode
 
 
-    @staticmethod
-    def render(render_root, pipe_url, pipe):
+    def render(self, render_root, pipe_url, pipe):
+        if not self.should_render:
+            return
         pdr = PipelineDepgraphRenderer(pipe)
```
karkhaz: Why construct a new `PipelineDepgraphRenderer` here? `render` is an instance method now. It should just call `self.render_to_file` instead of `pdf.render_to_file`. If you remove the `pipe` instance variable, you could add a `pipe` argument to `render_to_file` and pass it in.

lib/litani_report.py
```diff
@@ -93,8 +99,9 @@ def render_to_file(self, out_file):
         return not proc.returncode
 
 
-    @staticmethod
-    def render(render_root, pipe_url, pipe):
+    def render(self, render_root, pipe_url, pipe):
```
karkhaz, line 98: The `render_to_file` method above no longer needs to wrap `dot` in a try-catch because you're already checking that `dot` is installed in the constructor.

litani
```diff
@@ -668,11 +669,13 @@ async def run_build(args):
     report_dir = litani.get_report_dir()
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    litani_report.render(run, report_dir)
+    pipeline_degraph_renderer = litani_report.PipelineDepgraphRenderer(pipe=None)
```
karkhaz: No need for `pipe=None` if you remove the instance variable.

karkhaz (changes requested):

karkhaz (approved):

Issue #44, opened 2021-09-20
Add description and command to pipeline dependency graph command nodes


karkhaz:
Currently, the pipeline dependency graph is not as useful as it could be. The file nodes contain a truncated file name, and the command nodes contain a truncated description.

Acceptance:
* Use [Graphviz HTML-like labels](https://graphviz.org/doc/info/shapes.html#html) to create a two-section node for commands. The top section should contain the description (untruncated), and the bottom section should contain the full command line.

PR #56, opened 2021-10-08
Updated dependency graph command nodes to have pipeline_name, description and command

ronakfof:
*Issue #, if available:*
https://github.com/awslabs/aws-build-accumulator/issues/44

*Description of changes:*
Added description and command to pipeline dependency graph command nodes

*Example*
Screen Shot 2021-10-08 at 6 14 26 PM

### DIFF
--- a/lib/graph.py
+++ b/lib/graph.py
@@ -73,7 +73,8 @@ class CommandNode(Node):
         self.command = '<BR/>'.join(textwrap.wrap(command, width=line_width))
         self.style = style
 
-        self.style["shape"] = "box"
+        self.style["shape"] = "plain"
+
 
 
     def __hash__(self):
@@ -85,11 +86,21 @@ class CommandNode(Node):
 
 
     def __str__(self):
-        return '"{id}" [{style}];'.format(
-            id=self.id, style=",".join([
+        return '''"{id}" [label=<
+            <TABLE BORDER="0" CELLBORDER="1" CELLSPACING="0">
+                <TR>
+                    <TD><B>{pipeline_name}</B></TD>
+                    <TD><B>{description}</B></TD>
+                </TR>
+                <TR>
+                    <TD COLSPAN="2">{command}</TD>
+                </TR>
+            </TABLE>> {style}];'''.format(
+            id=self.id, description = self.description,
+            command = self.command, pipeline_name = self.pipeline_name,
+            style=",".join([
                 f'{key}="{Node.escape(value)}"'
-                for key, value in self.style.items()
-            ]))
+                for key, value in self.style.items()]))
 
 
 
@@ -131,11 +142,9 @@ class SinglePipelineGraph:
     def build(self):
         for job in self.iter_jobs():
             args = job["wrapper_arguments"]
-            cmd_label = args.get("description", args["command"])
-            if not cmd_label:
-                cmd_label = args["command"]
             cmd_node = self._make_cmd_node(
-                job["complete"], job.get("outcome", None), cmd_label)
+                job["complete"], job.get("outcome", None),
+                args["pipeline_name"], args["description"], args["command"])
             self.nodes.add(cmd_node)
 
             for inputt in args.get("inputs") or []:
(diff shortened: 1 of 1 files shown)


lib/graph.py
```diff
@@ -34,6 +34,11 @@ def escape(string):
         return string
 
 
+def wrap_command(limit, break_with, command):
+    occurrences = re.findall('.{1,%i}' % limit, command)
```
karkhaz: I think it would be better to use [`textwrap.wrap`](https://docs.python.org/3/library/textwrap.html#textwrap.wrap) or the class version, as it is more likely to end the line on a hyphen or whitespace

lib/graph.py
```diff
@@ -46,7 +51,11 @@ def __init__(self, fyle, truncate=30, **style):
         self.style = style
 
         if len(path.name) + len(ext) > truncate:
-            self.style["label"] = path.name[:(truncate - 1 - len(ext))] + "…" + ext
+            self.style["label"] = wrap_command(
```
karkhaz: Please make this consistent with the surrounding style, this codebase doesn't have newlines after each parameter. Suggest:
```python
path_name = wrap_command(truncate - 1 - len(ext), '\n', path.name)
self.style["label"] = f"{path_name}{ext}"
```

karkhaz: The change to split the filenames over multiple lines is different to having a description + command in the command box, please make it a separate commit

lib/graph.py
```diff
@@ -34,6 +34,11 @@ def escape(string):
...
 
 
+def wrap_command(limit, break_with, command):
+    occurrences = re.findall('.{1,%i}' % limit, command)
+    newline_separated_command = break_with.join(occurrences)
+    return newline_separated_command
+
 
 class DependencyNode(Node):
     def __init__(self, fyle, truncate=30, **style):
```
karkhaz: Suggest changing the `truncate` argument to `line_width`, since we're no longer truncating.

lib/graph.py
```diff
@@ -66,18 +75,14 @@ def __str__(self):
                 for key, value in self.style.items()]))
 
 
-
 class CommandNode(Node):
-    def __init__(self, command, truncate=30, **style):
-        self.command = command
+    def __init__(self, description, command, truncate=40, **style):
```
karkhaz: Suggest changing the `truncate` argument to `line_width`, since we're no longer truncating.

lib/graph.py
```diff
@@ -66,18 +75,14 @@ def __str__(self):
...
         self.id = hash(command)
+        self.description = wrap_command(truncate-1, '<BR/>', description)
```
karkhaz: Suggest to also make the description bold, so that it's better distinguished from the command?

lib/graph.py
```diff
@@ -135,11 +149,12 @@ def iter_jobs(self):
     def build(self):
         for job in self.iter_jobs():
             args = job["wrapper_arguments"]
-            cmd_label = args.get("description", args["command"])
-            if not cmd_label:
-                cmd_label = args["command"]
+            description = args["description"]
```
karkhaz: These variables are only used once. I would just write `args["description"]` in the function call.

lib/graph.py
```diff
@@ -219,12 +234,11 @@ def __str__(self):
 
         for job in self.iter_jobs():
             args = job["wrapper_arguments"]
+            
+            description = args["description"]
```
karkhaz: No need, just `cmd_node = CommandNode(args["description"], args["command"])`

karkhaz (changes requested):
Nice, thanks for doing this!

nrdg42 (approved):

lib/graph.py
```diff
@@ -166,7 +173,9 @@ def _make_cmd_node(complete, outcome, command):
             raise RuntimeError("Unknown outcome '%s'" % outcome)
         else:
             cmd_style["fillcolor"] = "#eceff1"
-        return lib.graph.CommandNode(command, **cmd_style)
+        return lib.graph.CommandNode(
+            pipeline_name, description, command, **cmd_style
```
karkhaz: Could we retain the current style here? The `)` doesn't need to be on a separate line. Same below.

karkhaz (changes requested):
Just one tiny comment :)

When this is done, I'll merge, as this should really be two separate PRs. I'm going to merge and then we can review what the git tree looks like, just to get an idea of how merging works so that you can also do it.

PR #57, opened 2021-10-14
Add summary row to HTML dashboard

karkhaz:
This PR adds a summary row to the HTML dashboard. This includes the following information:
* A summary of the number of failed, in-progress, passed, and total number of pipelines.
* For each CI stage *C*, the total number of pipelines in which all of the jobs for *C* have completed, as well as the number of jobs in *C* that have completed successfully.

This is depicted here:

Screenshot 2021-10-14 at 01 07 35

This allows users to gain an at-a-glance understanding of the run's progress. While pipelines may be 'in progress', there is no equivalent notion for a CI stage; so a run can be considered as succeeding so far if all of the CI stages have an equal number for 'successful' and 'completed'.

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -513,6 +513,19 @@ def get_git_hash():
         return None
 
 
+def get_summary(run):
+    ret = {
+        "in_progress": 0,
+        "success": 0,
+        "total": 0,
+        "fail": 0,
+    }
+    for pipe in run["pipelines"]:
+        ret["total"] += 1
+        ret[pipe["status"]] += 1
+    return ret
+
+
 def render(run, report_dir, pipeline_depgraph_renderer):
     temporary_report_dir = litani.get_report_data_dir() / str(uuid.uuid4())
     temporary_report_dir.mkdir(parents=True)
@@ -540,7 +553,8 @@ def render(run, report_dir, pipeline_depgraph_renderer):
     page = dash_templ.render(
         run=run, svgs=svgs, litani_hash=get_git_hash(),
         litani_version=litani.VERSION,
-        litani_report_archive_path=litani_report_archive_path)
+        litani_report_archive_path=litani_report_archive_path,
+        summary=get_summary(run))
     with litani.atomic_write(temporary_report_dir / "index.html") as handle:
         print(page, file=handle)
 
--- a/templates/dashboard.jinja.html
+++ b/templates/dashboard.jinja.html
@@ -230,6 +230,27 @@ p {
 .graph {
   margin-bottom: 6em;
 }
+
+#pipeline-stats-container {
+  display: flex;
+  margin-left 3em;
+  justify-content: space-evenly;
+}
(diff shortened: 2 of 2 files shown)


lib/litani_report.py
```diff
@@ -511,6 +511,34 @@ def get_git_hash():
...
+        ret[pipe["status"]] += 1
+        for stage in pipe["ci_stages"]:
+            try:
+                ret["stages"][stage["name"]]
+            except KeyError:
+                ret["stages"][stage["name"]] = {
+                    "completed": 0,
+                    "success": 0,
+                }
+            if stage["progress"] != 100:
```
angelonakos: For a particular pipeline and stage, a job can be classified as _completed_ with a _progress_ of 100, yet its status can be any one of: `success`, `fail` and `fail_ignored`. 

Should we provide a further breakdown into the completed jobs? 
We could add more lines after _COMPLETED_ that list the numbers for `success`, `fail` and `fail_ignored`.
Summing up these 3 numbers should give the same number as the one being displayed next to the _COMPLETED_ label.

angelonakos: Furthermore, should statistics summarizing the number of jobs, whose _progress_ was not equal to 100, also be included in the "stage stats" `<div></div>`?

angelonakos: Just tested these changes. Here are 2 screenshots illustrating the changing state of the "summary statistics" for the same run. In the first screenshot, one of the pipelines is still in progress, whereas in the second screenshot all pipelines have finished execution.

Screen Shot 2021-10-14 at 11 44 08

image

In this first image, I notice that in the _BUILD_ column stage statistics mention _4 COMPLETED_ and _4 SUCCESSFUL_ (this refers to the 4 pipelines whose status bar is 🟦 ). However, a statistic is not being reflected for `s2n_stuffer_read` which experienced a failure during the _build_ stage. Here's a more detailed view into this pipeline's _build_ stage:

image

**Question 1:** Should a statistic be captured and displayed to reflect what happened during the _build_ stage of `s2n_stuffer_read`?

----

Additionally, you can see that in the _TEST_ column in both images, the stage statistics mention _3 COMPLETED_ and _2 SUCCESSFUL_.  

**Question 2:** Should a statistic be captured and displayed to reflect the _number of jobs that were not attempted_ due to errors that occurred in some previous stage? I'm looking at the `s2n_stuffer_read` pipeline, which essentially has "1 job not attempted" both during the _test_ and the _report_ stages. Also, the `s2n_connection_get_handshake_type_name` pipeline has "1 job not attempted" during the _report_ stage. In the report stage, I would expect to see a number reflecting the fact that a total of 2 jobs were not attempted.

karkhaz:
Due to a contentious offline discussion, I've removed the second commit from this PR until we can spend some more time on it. The PR now includes only the pipeline headers, rather than the CI stage headers.

angelonakos (approved):

PR #60, opened 2021-10-19
Add release script

karkhaz:
This commit adds a script that creates a new release when run. It takes the following actions:

- Update the version number in lib/litani.py;
- Generate a changelog and prompt the user to edit it;
- Merge the release branch into develop, using the changelog for the merge commit;
- Tag the release;
- Create a new release candidate on the develop branch.

### DIFF
--- a/release
+++ b/release
@@ -1,0 +1,217 @@
+#!/bin/sh
+# vim: syntax=make ft=make
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at #
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+
+# This file is a shell script that writes the bottom half of itself to a
+# Makefile, and then runs that Makefile with make. It maintains state in the
+# tmp/litani-release directory, so you can read the files in there to help with
+# debugging.
+
+MAKEFILE_START_LINE=$(\
+  grep -ne '^# 8><' "$0" \
+  | tail -n 1 \
+  | awk -F: '{print $1}')
+
+LITANI_RELEASE=tmp/litani-release
+mkdir -p "${LITANI_RELEASE}"
+
+MAKEFILE=${LITANI_RELEASE}/Makefile
+tail -n+${MAKEFILE_START_LINE} "$0" > "${MAKEFILE}"
+
+make -Bj 1 -f "${MAKEFILE}"
+SUCCESS="$?"
+
+exit "${SUCCESS}"
+
+# 8>< - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+#
+# Makefile starts here
+
+TMP=tmp/litani-release
+
+SANITY_CHECKS = \
+	check-upstream-origin \
+	check-branches-up-to-date \
+	check-tac-installed
+
+RELEASE_PROCESS = \
+	checkout-develop \
+	merge-develop-release \
+	commit-changelog \
+	tag-release \
+	create-rc
+
+STATE_FILES = \
+	$(TMP)/current-release-number \
+	$(TMP)/rc-number \
+	$(TMP)/commits.json
+
+create-release: $(SANITY_CHECKS) $(STATE_FILES) $(RELEASE_PROCESS)
+	$(info --- Finished)
+	git checkout release
+	@>&2 printf '\n```````````````````````````````````````````````````````````\n'
+	@>&2 printf \
+	  "Now run\n\n  git push -u origin release\n  git push origin --tags\n\n"
+	@>&2 printf '___________________________________________________________\n'
+
+.PHONY: $(SANITY_CHECKS) $(PHONY_DEPS)
+
+check-upstream-origin:
+	$(info --- Checking that 'origin' remote points to upstream repository)
+	git remote show -n origin \
+	  | grep -e 'Fetch URL:\s\+.\+awslabs/aws-build-accumulator'
+
+check-branches-up-to-date:
+	$(info --- Ensuring local branches track upstream)
+	[ $$(git rev-parse develop) = $$(git rev-parse origin/develop) ]
+	[ $$(git rev-parse release) = $$(git rev-parse origin/release) ]
+
+
+check-tac-installed:
+	$(info --- Testing for 'tac' installation)
+	printf '' | tac
+
+checkout-develop:
+	$(info --- Checking out develop)
+	git checkout develop
+
+create-rc: $(TMP)/rc-number
+	$(info --- Creating new release candidate on develop branching off release)
+	git checkout -B develop
+	new_minor=$$(cat $(TMP)/rc-number); \
+	  sed -E \
+	    -e "s/^VERSION_MINOR.+/VERSION_MINOR = $${new_minor}/g" \
+	    -I '' lib/litani.py
+	sed -E -e 's/^RC *= *False/RC = True/g' -I '' lib/litani.py
+	git add lib/litani.py
+	major=$$(grep -e '^VERSION_MAJOR' lib/litani.py | head -n 1 | awk -F= '{print $$2}'); \
+	minor=$$(cat $(TMP)/rc-number); \
+	  git commit -m "Create version $${major}.$${minor} release candidate"
+
+$(TMP)/rc-number:
+	$(info --- Generating RC number)
+	printf "%d\n" \
+	  $$(( $$(grep -e '^VERSION_MINOR' lib/litani.py | awk '{print $$3}') + 1 )) \
+	  > $@
+	cat $@
+
+$(TMP)/current-release-number:
+	$(info --- Generating current release number)
+	printf "%d.%d.%d\n" \
+	  $$(grep -e '^VERSION_MAJOR' lib/litani.py | awk '{print $$3}') \
+	  $$(grep -e '^VERSION_MINOR' lib/litani.py | awk '{print $$3}') \
+	  $$(grep -e '^VERSION_PATCH' lib/litani.py | awk '{print $$3}') \
+	  > "$@"
+	cat $@
+
+merge-develop-release: $(TMP)/current-release-number
+	$(info --- Merging develop into release branch)
+	git checkout release
+	git merge \
+	  --no-ff \
+	  --message="Bump version to $$(cat $(TMP)/current-release-number)" \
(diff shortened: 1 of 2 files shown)


release
```diff
@@ -0,0 +1,154 @@
...
+make -Bj 1 -f "${MAKEFILE}"
+SUCCESS="$?"
+
+exit "${SUCCESS}"
+
+# 8>< - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+#
+# Makefile starts here
+
+TMP=/tmp/litani-release
```
angelonakos: Change to `tmp/litani-release` ?

release
```diff
@@ -0,0 +1,154 @@
...
+
+create-release: \
+	$(TMP)/current-release-number \
```
angelonakos: Should a command be added to remove any previous artifacts present inside of `$(TMP)`  prior to creating a new release?

release
```diff
@@ -0,0 +1,154 @@
...
+
+$(TMP)/tag-body: $(TMP)/commits.json
+	jq -r '.[] | "- " + .subject' > $@ < $<
+
+$(TMP)/commits.json:
+	commit=HEAD; \
+	while true ; do \
+	  author=$$(git show --no-patch --pretty=format:%an "$${commit}"); \
+	  subject=$$(git show --no-patch --pretty=format:%s "$${commit}"); \
+	  body=$$(git show --no-patch --pretty=format:%b "$${commit}"); \
```
angelonakos, line 201: If HEAD contains one of `{ } [ ]` will the values of the `subject` or `body` keys cause the resulting JSON to be invalid?

release
```diff
@@ -0,0 +1,154 @@
...
+	git commit -m "Create version 1.$$(cat $(TMP)/rc-number) release candidate"
+
+$(TMP)/rc-number:
+	printf "%d\n" \
+	  $$(( $$(grep -e '^VERSION_MINOR' lib/litani.py | awk '{print $$3}') + 1 )) \
+	  > $@
+
+$(TMP)/current-release-number:
+	printf "%d.%d.%d\n" \
+	  $$(grep -e '^VERSION_MAJOR' lib/litani.py | awk '{print $$3}') \
```
angelonakos, line 113: Minor suggestion if we want to separate "important variables": `VERSION_MAJOR, VERSION_MINOR, VERSION_PATCH, RC` variables could be part of a  `lib/litani_version.py` file instead, and those values can then be imported inside of `lib/litani.py`

release
```diff
@@ -0,0 +1,154 @@
...
+
+create-rc: $(TMP)/rc-number
+	git checkout -B develop
+	new_minor=$$(cat $(TMP)/rc-number); \
+	  sed \
+	    -e "s/^VERSION_MINOR.+/VERSION_MINOR = $${new_minor}/g" \
+	    -I '' lib/litani.py
+	sed -e 's/^RC *= *False/RC = True/g' -I '' lib/litani.py
+	git add lib/litani.py
+	git commit -m "Create version 1.$$(cat $(TMP)/rc-number) release candidate"
```
angelonakos: Inside the commit message should `1` be replaced by `$(VERSION_MAJOR)`?

release
```diff
@@ -0,0 +1,154 @@
...
+	  grep -ne '^Version ' CHANGELOG \
+	  | head -n 1 \
+	  | awk -F: '{print $$1}' \
+	); \
+	tail -n+$${start_line} CHANGELOG > $@
+
+$(TMP)/user-changelog: $(TMP)/commits.json
+	printf "" > $@
+	printf "%s\n" "# This will be the changelog entry for this release." >> $@
+	printf "%s\n" "# Lines beginning with '#' will be ignored." >> $@
```
angelonakos, line 164: When I ran this locally, I left only one newline in the end of the user-changelog whereas I should have left two newlines in order for the updated-changelog to maintain the standard "two newlines" separating sections about different versions.

Should we add `	printf "%s\n" "# Leave 2 newlines following the last entry." >> $@` ?

release
```diff
@@ -0,0 +1,154 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+MAKEFILE_START_LINE=$(\
+  grep -ne '^# 8><' "$0" \
+  | tail -n 1 \
+  | awk -F: '{print $1}')
+
+LITANI_RELEASE=/tmp/litani-release
```
angelonakos: Change to `tmp/litani-release` ?

release
```diff
@@ -0,0 +1,154 @@
...
+mkdir -p "${LITANI_RELEASE}"
+
+MAKEFILE=$(mktemp ${LITANI_RELEASE}/XXXXXX || exit 1)
```
angelonakos: I prefer `MAKEFILE=litani-release-makefile` only because it's clear to someone after running the release script what the file with the unfamiliar name actually is.

release
angelonakos: I must note that `/tmp/litani-release` wasn't "cleaned up" when I initially tested this.

release
angelonakos: This suggestion doesn't have to be part of this PR. Just a future recommendation.

release
karkhaz: if you're talking about `$commit`: HEAD is literally the string `"HEAD"`, it's not a variable. This loop only ever appends the `~` character to the `$commit` variable, so it goes `HEAD`, `HEAD~`, `HEAD~~`, ...

If you're talking about the actual values of subject, body, etc then jq already deals with that as we're passing `--raw-input --slurp`. This makes JQ read the metadata for a single commit as one giant string; JQ isn't expecting the string to be escaped, it does that for us when it converts it to JSON internally.

angelonakos: You're right I should have referred to `$commit` which gets updated in each iteration of the loop to `HEAD,` `HEAD~`, `HEAD~~`. Thanks for pointing out that `--raw-input --slurp` handles escaping for us

release
karkhaz: > Should a command be added to remove any previous artifacts present inside of $(TMP) prior to creating a new release?

No need, these all get overwritten

karkhaz: > I must note that /tmp/litani-release wasn't "cleaned up" when I initially tested this.

This is intentional to help with debugging. I'm going to write a comment explaining which files to look in in case something goes wrong.

release
karkhaz: Not sure that this would be an advantage, we'd still need to parse them out with awk and friends if we're to use them from a shell script. I guess one small advantage is that if `VERSION` is defined in lib/litani.py and `VERSION_MINOR` is defined in lib/version.py, then there's no need to do `head` or `tail` to find the right occurrence of it...

release
karkhaz: I fixed the code so that it discards whatever blanks the user added at the beginning and end of the file, after stripping out comments, and then adds two more blanks.

release
```diff
@@ -0,0 +1,185 @@
...
+	tag-release \
+	create-rc
+
+tac-installed:
+	$(info --- Testing for 'tac' installation)
+	printf '' | tac
+
+checkout-develop:
+	$(info --- Checking out develop)
+	git checkout develop
```
angelonakos, line 88: I had to use: `git checkout --track origin/develop`

Should we change this line to that?

Initially, I got an error because my config is missing the following setting: `checkout.defaultRemote=origin` as mentioned in the hint by git below:

```bash
$ ./release
--- Testing for 'tac' installation
printf '' | tac
--- Checking out develop
git checkout develop
hint: If you meant to check out a remote tracking branch on, e.g. 'origin',
hint: you can do so by fully qualifying the name with the --track option:
hint:
hint:     git checkout --track origin/<name>
hint:
hint: If you'd like to always have checkouts of an ambiguous <name> prefer
hint: one remote, e.g. the 'origin' remote, consider setting
hint: checkout.defaultRemote=origin in your config.
fatal: 'develop' matched multiple (3) remote tracking branches
make: *** [checkout-develop] Error 128
```

karkhaz:
Thanks for all these comments. I've structured the list of commands into three logical sections, and the list of debug messages now reflects the structure of how this script works: do a bunch of sanity checks, write some data out to touch files, then run the actual release process.
```
--- Checking that 'origin' remote points to upstream repository
--- Ensuring local branches track upstream
--- Testing for 'tac' installation
--- Generating current release number
--- Generating RC number
--- Generating JSON list of feature commits
--- Checking out develop
--- Merging develop into release branch
--- Generating changelog for this release from feature list
--- Dumping old changelog
--- Assembling new changelog from user input
--- Turning off '-rc' suffix in version number
--- Adding changelog and updated version number
--- Generating tag body
--- Tagging release
--- Creating new release candidate on develop branching off release
```

release
karkhaz: I've added two new checks that run before the release process, the second one should hopefully address this comment:
1. Check that the remote you call `origin` actually is this GitHub awslabs repo
2. Check that the branch you call `develop` points to the same commit as `origin/develop`, and same for `release`.

One thing that it doesn't do is fetch from remote, I'm relying on the user to do that... maybe we should fetch just to be safe? But so far, this script works entirely locally and doesn't do anything with the internet at all, it relies on you to push also, and I think that's a nice property to have some how.

angelonakos: My `origin` was indeed the awslabs repo and prior to running the script my `develop` pointed to the same commit as `origin/develop` (same for `release`). Prior to running the script, I also ran `git fetch --all` just to retrieve updates from multiple remotes (awslabs, yours, mine, Ronak's and Nick's).

I agree that not reaching out either to fetch or to push is best. In the latter case, we avoid executions of this script accidentally overwriting some branch.

angelonakos (approved):

release
karkhaz: Are you absolutely sure that you had `develop` checked out when you ran this the first time? My understanding is that you wouldn't have got that error if you had an existing `develop` branch.

Would you mind running this script with a checked out `develop` pointing to `origin/develop` and try running this again? If it still doesn't work we shouldn't merge this

angelonakos: > Are you absolutely sure that you had develop checked out when you ran this the first time?

You're right. I hadn't checked out develop. What I had done: 

1. Clone the repo.
2. Add remotes for your, mine, Nick's and Ronak's forks.
3. Fetch from all remotes.
4. Checkout `kk-release-script`
5. Run `./release`

What I did just now (step #4 is was what I originally hadn't done before):

1. `git clone git@github.com:awslabs/aws-build-accumulator.git`
2. `cd aws-build-accumulator`
3. `git fetch`
4.  `git checkout develop` # Got this message: `branch 'develop' set up to track 'origin/develop'. Switched to a new branch 'develop'`
6. Added all remotes and then `git fetch --all`
7. `git checkout kk-release-script`
8. `./release`

karkhaz: Thanks. Then I think my addition this resolves this issue. The check
```
[ $$(git rev-parse develop) = $$(git rev-parse origin/develop) ]
```
will fail if step 4 had not been performed, or if for any other reason `develop` is pointing somewhere else.

nrdg42 (approved):

Issue #66, opened 2021-12-08
.litani_cache_dir discovery can cause run-cbmc-proofs.py failure.


oliverlavery:
Using Litani to build CBMC proofs we noticed an undesirable behaviour: because litani searches for .litani_cache_dir dotfiles in subdirectories of the cwd as well as parents, a cache dotfile left from a build in a subdirectory can cause a build in a parent directory to fail.

e.g.:

```
lavero in projdir/cbmc/proofs/load_locker_state at docker-hal on  remotes/foo/bar/main/dev [!+?]
➜ cat .litani_cache_dir
/tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d

lavero in projdir/cbmc/proofs/load_locker_state at docker-hal on  remotes/foo/bar/main/dev [!+?]
➜ cd ..

lavero in projdir/cbmc/proofs at docker-hal on  remotes/foo/bar/main/dev [!+?] via 🐍 system
➜ ls
load_locker_state  Makefile-project-defines  Makefile-project-testing   output     run-cbmc-proofs.py
Makefile.common    Makefile-project-targets  Makefile-template-defines  README.md

lavero in projdir/cbmc/proofs at docker-hal on  remotes/foo/bar/main/dev [!+?] via 🐍 system
➜ ./run-cbmc-proofs.py

For your convenience, the output of this run will be symbolically linked to /home/lavero/code/projdir/cbmc/proofs/output/latest/html/index.html

Configuring CBMC proofs: 0 / 1get_cache_dir called
_get_cache_dir returning: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d
get_cache_dir called
_get_cache_dir returning: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d
We're adding a job!: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d/jobs
get_cache_dir called
_get_cache_dir returning: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d
get_cache_dir called
_get_cache_dir returning: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d
get_cache_dir called
_get_cache_dir returning: /tmp/litani/runs/ba9fae78-441b-4d2a-b11e-ae18fbe4138d
[...]

PR #68, opened 2021-12-23
Fix .litani_cache_dir lookup

ronakfof:
*Issue #, if available:*
#66 

*Description of changes:*
Removed lookup for `.litani_cache_dir` in subdirectories

### DIFF
--- a/lib/litani.py
+++ b/lib/litani.py
@@ -154,10 +154,6 @@ def _get_cache_dir(path=os.getcwd()):
         while current.parent != current:
             current = current.parent
             yield current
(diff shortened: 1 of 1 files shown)


karkhaz:
Fine, thanks, could you use a better commit message?

IMHO a better title would be "Do not search for litani cache in subdirectories" or something.

The body should explain that prior to this commit, running Litani in a parent directory after running it in a child would cause unexpected behaviour, basically summarize the problem in #66.

At the bottom of the commit message, it should say "This commit fixes #66". That will make the issue auto-close once we merge this commit.

nrdg42 (approved):

karkhaz (approved):

Issue #67, opened 2021-12-23
Distinguish running from not-started jobs in HTML output


karkhaz:
The pipeline page should render jobs that have started (but not completed) differently from ones that have not yet started. Currently, users cannot know that a job has started from any of Litani's output, which makes it difficult to diagnose jobs that are hung, etc.

angelonakos:
The fix for this issue was released in release 1.17.0

PR #71, opened 2022-01-05
Show running jobs on pipeline page

nrdg42:
Distinguish the jobs that are currently running from the jobs that have
not started on the pipeline HTML page. Use CSS to color jobs that are
currently running orange on the dependency graph, border them in orange
on the pipeline page, and only show the in progress icon on running jobs
in the table of contents.

image

image

*Issue #, if available:*
#67  

*Description of changes:*

### DIFF
--- a/lib/graph.py
+++ b/lib/graph.py
@@ -172,8 +172,9 @@ class SinglePipelineGraph:
         for job in self.iter_jobs():
             args = job["wrapper_arguments"]
             cmd_node = self._make_cmd_node(
-                job["complete"], job.get("outcome", None),
-                args["pipeline_name"], args["description"], args["command"])
+                job["complete"], bool(job.get("start_time")), job.get("outcome",
+                None), args["pipeline_name"], args["description"],
+                args["command"])
             self.nodes.add(cmd_node)
 
             for inputt in args.get("inputs") or []:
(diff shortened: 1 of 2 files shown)


angelonakos (approved):

karkhaz (changes requested):
This is great, thank you!

I think the orange is too close to the colour we use for `fail_ignored`, which is also orange rather than red. Can I suggest to give the div a black border for running? I think that still distinguishes it reasonably well from not-started, which has an invisible border.

For the dependency graph, I suggest white node with black text for running, and white node with grey text for not started.

feliperodri:
@nrdg42 That's great!! Thanks for implementing this change. I agree with @karkhaz about the color pattern. We should try to use distinct and intuitive colors, but most important we need a **legend color**! New users need a straightforward way to know what each color represent.

nrdg42:
How about something like this for now? I see the need for a Legend like @feliperodri says.


image
image

karkhaz (approved):

karkhaz:
That's a good idea @feliperodri ... that should be reasonably easy to do, let's merge this and we'll put the legend on the backlog, shouldn't take too long.

feliperodri (approved):

PR #72, opened 2022-01-05
Add litani transform-jobs command

karkhaz:
This PR adds a new command, `litani transform-jobs`, thattransforms the list of added jobs before running the build. This allows users to modify Litani-based builds that they do not control, by first allowing the build to add jobs, then running this command to transform the jobs, then running the build. The PR also adds a working example to a new `examples` subdirectory in the codebase, to illustrate how the command would be used.

### DIFF
--- a/doc/src/man/litani-transform-jobs.scdoc
+++ b/doc/src/man/litani-transform-jobs.scdoc
@@ -1,0 +1,246 @@
+litani-transform-jobs(1) "" "Litani Build System"
+
+; Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+; Licensed under the Apache License, Version 2.0 (the "License").
+; You may not use this file except in compliance with the License.
+; A copy of the License is located at
+
+;     http://www.apache.org/licenses/LICENSE-2.0
+
+; or in the "license" file accompanying this file. This file is distributed
+; on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+; express or implied. See the License for the specific language governing
+; permissions and limitations under the License.
+
+
+# NAME
+
+litani transform-jobs - Print a list of jobs to be modified before running
+
+
+# DESCRIPTION
+
+This program allows clients to add to, delete from, or modify a list of Litani
+jobs before running them. The program prints out all the jobs that have been
+added to the current run so far, and then expects to read a new list of jobs on
+stdin. This new list of jobs will be 'saved' and run upon invocation of
+*litani-run-build(1)*.
+
+This program prints the list of jobs that have been added so far in JSON format.
+It also expects to read a list of jobs as a JSON list. Each job in the list will
+follow the same schema as the elements of the *["jobs"]["wrapper_arguments"]*
+key in the litani _run.json_ file; see *litani-run.json(5)* for the schema. Most
+users will want to run this program before running *litani run-build*, which
+means that none of the jobs will have started; that means that the JSON schema
+will match that of an unstarted job, meaning the *complete* key will be _false_
+and there will be no *start_time* key.
+
+If this program prints a job, and the JSON list written to stdin contains a job
+with the same job id, then Litani will overwrite the original job with the one
+given on stdin. The 'job id' is the value of *["job_id"]* in the job dictionary,
+as specified in *litani-run.json(5)*.
+
+If this program prints a job, but a job with the same job id does not exist in
+the list of jobs written to this program's stdin, then that job will be deleted.
+
+If this program reads a job on stdin whose job id did not exist in the list of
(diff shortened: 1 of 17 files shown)


nrdg42 (approved):

lib/transform_jobs.py
```diff
@@ -0,0 +1,124 @@
...
+            j for j in user_jobs
+            if j["job_id"] not in self.old_uuids]
+        for job in new_jobs:
+            self._write_new_job(job)
+
+
+    @staticmethod
+    def _job_name(job):
+        desc = f" ({job['description']})" if job["description"] else ''
+        return f"{job['job_id']}{desc}"
```
angelonakos, line 67: A job always has a description, right? If yes, the line could change to this:

`        return f"{job['job_id']}{{job['description']}}"`

https://github.com/awslabs/aws-build-accumulator/blob/release/litani#L478

lib/transform_jobs.py
```diff
@@ -0,0 +1,124 @@
...
+    sys.stdout.flush()
+    os.close(sys.stdout.fileno())
+
+
+def _read_jobs():
+    in_text = sys.stdin.read()
+    return json.loads(in_text)
+
+
+async def main(_):
```
angelonakos, line 111: Is `_` needed?

doc/src/man/litani-transform-jobs.scdoc
```diff
@@ -0,0 +1,153 @@
+litani-transform-jobs(1) "" "Litani Build System"
```
angelonakos, line 1: also, add new Litani capability here: https://github.com/awslabs/aws-build-accumulator/blob/release/doc/src/man/litani.scdoc#L87

lib/transform_jobs.py
karkhaz: ...I don't even remember writing that line, and I'm strongly tempted to get rid of it. That's just nonsense. We should not synthesize a description if the user didn't provide one.

The schema says that description is optional: https://awslabs.github.io/aws-build-accumulator/#Schema_for_'wrapper_arguments'_key and that was always my intent. I do think it's important for users to be able to 'try out' Litani commands in a lightweight way before writing a proper run script, so I was trying very hard to make everything optional, even `description` which you almost certainly want to have when writing the run script for real.

I'm going to keep the code here as it is.

lib/transform_jobs.py
karkhaz: Yes. The way the argument parser works is that for each subcommand (`init`, `add-jobs`, ...), we associate the subcommand with a function that gets called. For example, if the user runs `litani init`, then the `init` function gets called because of [this line](https://github.com/awslabs/aws-build-accumulator/blob/release/litani#L206)
```py
init_pars.set_defaults(func=init)
```

All of these functions that serve as command entrypoints have the same signature: they take a single argument, which is the `args` object itself---that happens [here](https://github.com/awslabs/aws-build-accumulator/blob/release/litani#L839):
```
await args.func(args)
```
which I admit is a bit meta and obtuse :) Anyway, the point is that for `transform_jobs`, we set the entry point to `lib.transform_jobs.main` and pass the `args` object to it, but since this command has no command-line arguments, the `main` method doesn't use the object that it was given. So we do need the argument, but I named it `_` to make it clear that it's unused (and to shut pylint up).

doc/src/man/litani-transform-jobs.scdoc
karkhaz: Done, thank you!

angelonakos (approved):

examples/no-standalone-transform/run-all.py
```diff
@@ -0,0 +1,58 @@
...
+        subprocess.run([fyle, "--no-standalone"], check=True)
+
+
+# ``````````````````````````````````````````````````````````````````````````````
+# IMPORTANT! bufsize must be 0, otherwise the C stdlib will buffer the jobs that
+# we send back to Litani's stdin
+#
+proc = subprocess.Popen(
+    [litani, "-v", "transform-jobs"], stdin=subprocess.PIPE,
+    stdout=subprocess.PIPE, text=True, bufsize=0)
```
markrtuttle, line 45: Could you include an example of something on the stdin?

test/e2e/run
```diff
@@ -79,6 +79,25 @@ def add_jobs(litani, run_dir, mod):
             litani, "add-job", *job.get("args", []), **job.get("kwargs", {}))
 
 
+def transform_jobs(litani, run_dir, mod):
```
markrtuttle, line 82: Confused by the asynchrony here.  Don't follow what the end-to-end stuff is doing.

markrtuttle: Resolved during discussion with Kareem.

markrtuttle (approved):

PR #78, opened 2022-01-12
Add optional --output-dir flag to Litani test suite

### LINKED ISSUE #74: Allow users to specify an output directory for the Litani test suite
Currently, the Litani test suite writes its output to a file within the source directory, which is .gitignored. However, if the source directory has been installed in a root-owned location using a package manager, then running the tests will fail as it will not be possible to write the test directory.

Acceptance:
The `test/run` script should accept an optional directory as a command line argument. The test suite should not write any file outside that directory.

ronakfof:
Fixes #74

*Issue #, if available:*
#74 

*Test changes:*
`./test/run` [DEFAULT BEHAVIOR]
OR
`./test/run --output-dir /tmp/litani-results/` [ADDED OPTIONAL FLAG]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 test/output
-test/e2e/output
 
 .litani_cache_dir
 
--- a/test/run
+++ b/test/run
@@ -14,6 +14,7 @@
 # permissions and limitations under the License.
 
 
+import argparse
 import logging
 import os
 import pathlib
@@ -21,6 +22,21 @@ import re
 import subprocess
 import sys
 import uuid
+
+DESCRIPTION = "Execute e2e and unit tests for Litani"
+
+
+def get_args():
+    pars = argparse.ArgumentParser(description=DESCRIPTION)
+    for arg in [{
+            "flags": ["--output-dir"],
+            "help": "output dir for test results",
+            "default": pathlib.Path(__file__).resolve().parent / "output",
+            "type": pathlib.Path
+    }]:
+        flags = arg.pop("flags")
+        pars.add_argument(*flags, **arg)
+    return pars.parse_args()
 
 
 def run_cmd(cmd):
@@ -53,7 +69,7 @@ def collapse(string):
     return re.sub(r"\s+", " ", string)
 
 
-def add_e2e_tests(litani, test_dir, root_dir, counter):
+def add_e2e_tests(litani, test_dir, counter, output_dir):
     e2e_test_dir = test_dir / "e2e"
     # 4 jobs per test (init, add-jobs, run-build, check-run)
     # skip __init__.py and __pycache__
@@ -63,7 +79,7 @@ def add_e2e_tests(litani, test_dir, root_dir, counter):
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
 
-        run_dir = e2e_test_dir / "output" / str(uuid.uuid4())
+        run_dir = output_dir / "e2e_outputs" / str(uuid.uuid4())
 
         litani_add(
             litani, counter,
@@ -143,16 +159,21 @@ def print_counter(counter):
 
 
 def main():
+    args = get_args()
     logging.basicConfig(format="run-tests: %(message)s")
     test_dir = pathlib.Path(__file__).resolve().parent
     root = test_dir.parent
     litani = root / "litani"
 
+    output_dir = args.output_dir.resolve()
+    output_dir.mkdir(exist_ok=True, parents=True)
+    os.chdir(output_dir)
+
     run_cmd([
         litani, "init",
         "--project", "Litani Test Suite",
-        "--output-prefix", test_dir / "output",
-        "--output-symlink", test_dir / "output" / "latest"])
+        "--output-prefix", ".",
+        "--output-symlink", "latest"])
 
     counter = {
         "added": 0,
@@ -160,7 +181,8 @@ def main():
     }
 
     add_unit_tests(litani, test_dir, root, counter)
-    add_e2e_tests(litani, test_dir, root, counter)
+    add_e2e_tests(
+        litani, test_dir, counter, output_dir=output_dir)
     print()
 
     run_cmd([litani, "run-build"])
--- a/test/README
+++ b/test/README
@@ -4,3 +4,8 @@
 The `run` script uses litani to run a variety of unit and system tests.
 The dashboard showing the output of the latest test run will always be
 symlinked from test/output/latest/html/index.html.
+
+# OPTIONS
+
+*--output-dir*
+	Litani will write all of its test output files for this run to this path


test/run
```diff
@@ -143,24 +161,29 @@ def print_counter(counter):
...
+        "--output-symlink", output_dir / "latest"])
 
     counter = {
         "added": 0,
         "total": 0,
     }
 
     add_unit_tests(litani, test_dir, root, counter)
-    add_e2e_tests(litani, test_dir, root, counter)
+    add_e2e_tests(litani, test_dir, root, counter, output_dir=output_dir)
```
angelonakos: `output_dir` will always have a value due to L169-172. `output_dir` should be part of this function's `*args`.

test/run
```diff
@@ -53,7 +68,7 @@ def collapse(string):
     return re.sub(r"\s+", " ", string)
 
 
-def add_e2e_tests(litani, test_dir, root_dir, counter):
+def add_e2e_tests(litani, test_dir, root_dir, counter, output_dir=None):
```
angelonakos: `output_dir` should be part of this function's `*args` and not be part of  `**kwargs`.

test/run
```diff
@@ -63,7 +78,10 @@ def add_e2e_tests(litani, test_dir, root_dir, counter):
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
 
-        run_dir = e2e_test_dir / "output" / str(uuid.uuid4())
+        if output_dir:
```
angelonakos: Right now, won't `output_dir` always have a value?

test/run
```diff
@@ -53,7 +68,7 @@ def collapse(string):
     return re.sub(r"\s+", " ", string)
 
 
-def add_e2e_tests(litani, test_dir, root_dir, counter):
+def add_e2e_tests(litani, test_dir, root_dir, counter, output_dir):
```
angelonakos: `root_dir` is unused. Can be removed

test/run
ronakfof: Resolved

test/run
ronakfof: Resolved

test/run
ronakfof: Resolved

test/run
ronakfof: Resolved

test/run
ronakfof: Resolved

test/run
```diff
@@ -143,24 +161,30 @@ def print_counter(counter):
...
 
     counter = {
         "added": 0,
         "total": 0,
     }
 
     add_unit_tests(litani, test_dir, root, counter)
-    add_e2e_tests(litani, test_dir, root, counter)
+    add_e2e_tests(
+        litani, test_dir, counter, output_dir=args.output_dir or None)
```
angelonakos: Why not use the value of `output_dir`?

ronakfof: Because in `add_e2e_tests()` there is no use of `test_dir / "output"` which is the alternate value when `args.output_dir` is not passed. Reason it is not used is because the default run dir for e2e tests is created under `test_dir / "e2e" / "output`

test/run
```diff
@@ -143,24 +161,30 @@ def print_counter(counter):
...
 def main():
+    args = get_args()
     logging.basicConfig(format="run-tests: %(message)s")
     test_dir = pathlib.Path(__file__).resolve().parent
     root = test_dir.parent
     litani = root / "litani"
+    if args.output_dir:
+        output_dir = pathlib.Path(args.output_dir)
+    else:
+        output_dir = test_dir / "output"
```
nrdg42: Why don't you give `output_dir` a default value in `get_args()`?

test/run
nrdg42: If I were the user and I specified `--output_dir <my-directory>`, I would expect the behavior to be either:
* that the tests that currently end up in `test_dir / "output"` to end up in `<my-directory> / "output"` and the tests that end up in `test_dir / "e2e" / "output"` to end up in `<my-directory> / "e2e" / "output"`

or
* that the tests that currently end up in `test_dir / "output"` to end up in `<my-directory>` and the tests that end up in `test_dir / "e2e" / "output"` to end up in `<my-directory> / "e2e" / "output"`

nrdg42:
Could you add a commit message explaining your changes?

test/run
ronakfof: Behavior currently matches the first one - 

> that the tests that currently end up in `test_dir / "output"` to end up in `<my-directory> / "output"` and the tests that end up in `test_dir / "e2e" / "output"` to end up in `<my-directory> / "e2e" / "output"`

test/run
karkhaz: Agree

test/run
karkhaz: The first one is fine. The most important thing about this PR is that when `--output-dir` is passed, there should be no files written outside that directory. That includes not only the top-level Litani test process, but also all of the Litani invocations that are part of each e2e test.

The reason for this PR is that if we install Litani in a root-owned location, we want to be able to run the test suite and guarantee that it won't try to write output files under the root-owned location where Litani is installed. So if we run the test suite with `--output-directory /tmp/litani`, absolutely everything needs to be written under there.

Probably the easiest way to try this out is just to try it, `sudo chown -R root: litani_checkout` and then try running the test suite (without sudo) and seeing if anything breaks.

karkhaz (changes requested):

angelonakos (approved):

test/run
```diff
@@ -143,24 +161,32 @@ def print_counter(counter):
...
     logging.basicConfig(format="run-tests: %(message)s")
     test_dir = pathlib.Path(__file__).resolve().parent
     root = test_dir.parent
     litani = root / "litani"
+    if args.output_dir:
+        output_dir = pathlib.Path(args.output_dir) / str(uuid.uuid4())
+        output_dir.mkdir(exist_ok=True, parents=True)
+        os.chdir(output_dir)
+    else:
+        output_dir = test_dir / "output"
```
nrdg42: It appears as though my previous comment was not addressed. Could you please explain? I believe that `args.output_dir` should have a default value of `test_dir / "output"` and when a user specifies a different output directory, that should be set as the output directory. Why do we create a need to make a subdirectory with a UUID for that directory?

nrdg42 (commented):
Could you please add a body to your commit message explaining your changes and why they are necessary?

test/run
ronakfof: Resolved

.gitignore
```diff
@@ -1,5 +1,4 @@
 test/output
-test/e2e/output
```
karkhaz, line 2: I think this should stay, because this is still the default location

test/run
```diff
@@ -22,6 +23,20 @@ import subprocess
...
 
+DESCRIPTION = "Execute e2e and unit tests for Litani"
+
+
+def get_args():
+    pars = argparse.ArgumentParser(description=DESCRIPTION)
+    for arg in [{
+            "flags": ["--output-dir"],
+            "help": "output dir for test results",
+            "default": pathlib.Path(__file__).resolve().parent / "output"
```
karkhaz: If the user supplies a value, it would be a `str`. But the default value you're giving here is a `pathlib.Path`. I don't like it when you're not sure what type an argument is, especially in cases like this: we can use the `/` operator on the output directory if it's a `Path` but not a `str`, for example.

Suggestions:
```
"default": str(pathlib.Path(__file__).resolve().parent / "output")
```
so you know that it will be a `str` either way, or
```
"type": pathlib.Path
```
so that you know it's a `Path` either way. (See [this](https://docs.python.org/3/library/argparse.html#type) for how that works)

test/run
```diff
@@ -143,24 +158,32 @@ def print_counter(counter):
...
 def main():
+    args = get_args()
     logging.basicConfig(format="run-tests: %(message)s")
     test_dir = pathlib.Path(__file__).resolve().parent
     root = test_dir.parent
     litani = root / "litani"
+    output_dir = pathlib.Path(args.output_dir).resolve()
+
+    temp = output_dir
+    while not temp.exists():
```
karkhaz: I don't understand what this is for?

I think if `output_dir` doesn't exist, we should just create it

karkhaz (changes requested):

test/run
ronakfof: `litani --init --output-prefix` does that for us.

Also if we create dir here, `chdir` into that and run `litani --init`, it will again create the same directories.
so the end result can be - `/tmp/litani-results/tmp/litani-results/...`

karkhaz: Ah right. That's because your changing into `temp` and then setting `--output-prefix` to `output_dir`. I just noticed also that if the user passes a string, that would be a relative directory, but the default value is an absolute path 🍄.

Could you try: get rid of the loop, `chdir` into `args.output_directory`, and then
```
--output-prefix .
--output-symlink latest
```

test/run
```diff
@@ -22,6 +23,21 @@ import subprocess
...
+DESCRIPTION = "Execute e2e and unit tests for Litani"
+
+
+def get_args():
+    pars = argparse.ArgumentParser(description=DESCRIPTION)
+    for arg in [{
+            "flags": ["--output-dir"],
+            "help": "output dir for test results",
+            "default": pathlib.Path(__file__).resolve().parent / "output",
+            "type": pathlib.Path
```
karkhaz, line 35: 🍄 This is going to give inconsistent results because the user might pass a relative path, but the default is an absolute path. I think we should `resolve` whatever the user gives us.

Thus, instead of `"type": pathlib.Path`, could you write a function that takes a string and returns a resolved path? Then you would do `"type": resolved_path`. Again, check out the 'type' documentation to see another example.

karkhaz (changes requested):

test/run
ronakfof: I resolve what user gives us on line `167` - 
```
output_dir = args.output_dir.resolve()
```

ronakfof:
```
ronakf litani % ./test/run --output-dir /Users/ronakf/Desktop/litani/tmp/litani-results
Report will be rendered at file:///Users/ronakf/Desktop/litani/tmp/litani-results/latest/html/index.html
89 / 89 tests added
[93/93] multiproc_dump_run: check run 
```

.gitignore
ronakfof: It's not anymore. `e2e` folder will be now created under `test/output`

karkhaz (approved):

angelonakos (approved):

PR #79, opened 2022-01-13
Add a --fast option to the Litani test suite

### LINKED ISSUE #76: Add a --fast option to the Litani test suite
Acceptance:
- Each Litani end-to-end test should have a module-level variable called SLOW, set to either True or False
- The test runner should not run SLOW tests if --fast is passed

ronakfof:
- Each Litani end-to-end test has a module-level variable
  called SLOW, set to either True or False
- The test runner should not run SLOW tests if --fast is passed

Fixes #76

*Issue #, if available:*
#76 
*Test changes:*
`./test/run` for default behavior
OR
`./test/run --fast` for running tests with `SLOW=False`

Each test is made up of 4 jobs: init, add-jobs, run-build, and check-run
If --fast flag is passed, the jobs are given the a timeout=10s
For each test, SLOW is set to True is jobs fail to pass the above timeout.

### DIFF
--- a/test/README
+++ b/test/README
@@ -9,3 +9,13 @@ symlinked from test/output/latest/html/index.html.
 
 *--output-dir*
 	Litani will write all of its test output files for this run to this path
+
+*--fast*
+    Do not run tests with SLOW variable set to True
+
+## Writing Tests
+
+Each test is made up of 4 jobs: init, add-jobs, run-build, and check-run
+If --fast flag is passed, each job is given a timeout of 10 seconds.
+In the worst case, a test would take 40 seconds to run.
+For each test, SLOW is set to True if jobs fail to pass the above timeout.
--- a/test/run
+++ b/test/run
@@ -15,6 +15,7 @@
 
 
 import argparse
+import importlib
 import logging
 import os
 import pathlib
@@ -33,7 +34,11 @@ def get_args():
             "help": "output dir for test results",
             "default": pathlib.Path(__file__).resolve().parent / "output",
             "type": pathlib.Path
-    }]:
+        }, {
+            "flags": ["--fast"],
+            "help": "run fast tests only",
+            "action": "store_true"
+        }]:
         flags = arg.pop("flags")
         pars.add_argument(*flags, **arg)
     return pars.parse_args()
@@ -45,6 +50,14 @@ def run_cmd(cmd):
     except subprocess.CalledProcessError:
         logging.error(
             "Failed to run command '%s'", " ".join([str(c) for c in cmd]))
+        sys.exit(1)
+
+
+def is_slow_test(module_file):
+    try:
+        return importlib.import_module(str(module_file.stem)).SLOW
+    except AttributeError:
+        logging.error("Variable SLOW is missing from: %s", module_file.name)
         sys.exit(1)
 
 
@@ -69,12 +82,12 @@ def collapse(string):
     return re.sub(r"\s+", " ", string)
 
 
-def add_e2e_tests(litani, test_dir, counter, output_dir):
+def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
     e2e_test_dir = test_dir / "e2e"
     # 4 jobs per test (init, add-jobs, run-build, check-run)
     # skip __init__.py and __pycache__
     counter["total"] += (len(os.listdir(e2e_test_dir / "tests")) - 2) * 4
-
+    sys.path.insert(1, str(e2e_test_dir / "tests"))
     for test_file in (e2e_test_dir / "tests").iterdir():
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
@@ -85,8 +98,12 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
                 if line.strip().startswith("def transform_jobs("):
                     add_transform_jobs = True
                     break
+        if fast and is_slow_test(test_file):
+            continue
 
         run_dir = output_dir / "e2e_outputs" / str(uuid.uuid4())
+
+        timeout=10 if fast else 0
 
         litani_add(
             litani, counter,
@@ -100,7 +117,8 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
             ci_stage="test",
             description=f"{test_file.stem}: init",
             outputs=run_dir / ".litani_cache_dir",
-            cwd=run_dir)
+            cwd=run_dir,
+            timeout=timeout)
 
         run_build_input = str(uuid.uuid4())
         litani_add(
@@ -116,7 +134,9 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
             description=f"{test_file.stem}: add jobs",
             inputs=run_dir / ".litani_cache_dir",
             phony_outputs=run_build_input,
-            cwd=run_dir)
+            outputs=f"{run_dir}/output/jobs",
+            cwd=run_dir,
+            timeout=timeout)
 
         if add_transform_jobs:
             add_jobs_output = run_build_input
@@ -149,7 +169,8 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
             description=f"{test_file.stem}: run build",
             inputs=run_build_input,
             outputs=f"{run_dir}/output/run.json",
-            cwd=run_dir)
+            cwd=run_dir,
+            timeout=timeout)
 
         litani_add(
             litani, counter,
@@ -163,7 +184,8 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
             ci_stage="report",
             description=f"{test_file.stem}: check run",
             inputs=f"{run_dir}/output/run.json",
-            cwd=run_dir)
+            cwd=run_dir,
+            timeout=timeout)
 
 
 def add_unit_tests(litani, test_dir, root_dir, counter):
@@ -186,7 +208,7 @@ def print_counter(counter):
 
 def main():
     args = get_args()
-    logging.basicConfig(format="run-tests: %(message)s")
+    logging.basicConfig(format="\nrun-tests: %(message)s")
     test_dir = pathlib.Path(__file__).resolve().parent
     root = test_dir.parent
     litani = root / "litani"
@@ -208,7 +230,7 @@ def main():
 
     add_unit_tests(litani, test_dir, root, counter)
     add_e2e_tests(
-        litani, test_dir, counter, output_dir=output_dir)
+        litani, test_dir, counter, output_dir, args.fast)
     print()
 
     run_cmd([litani, "run-build"])
--- a/test/e2e/tests/custom_stages.py
+++ b/test/e2e/tests/custom_stages.py
@@ -1,4 +1,5 @@
 NUM_STAGES = 4
+SLOW = False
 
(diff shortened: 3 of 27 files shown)


angelonakos (changes requested):
Could you please elaborate on how you determined which tests were "slow" and which ones were "fast"? What's the criterion/definition for slow and fast?

If a developer adds new tests to Litani, how will they determine if their tests are slow or fast?

ronakfof:
We have 4 jobs per test (`init`, `add-jobs`, `run-build`, `check-run`) - 
If run time for all these 4 jobs for a test is `<= 14s`, then `SLOW=False` for that test else `SLOW=True`

angelonakos:
> We have 4 jobs per test (`init`, `add-jobs`, `run-build`, `check-run`) - If run time for all these 4 jobs for a test is `<= 10s`, then `SLOW=False` for that test else `SLOW=True`

This "10 second" rule makes sense. Can you document it in test/README.md ?

Out of curiosity, when determining if something is slow or fast, should we run the test `N` times and then compute the mean duration time of a test after `N` runs? Or is running them once enough?

angelonakos (approved):

nrdg42:
I think it would be good to have some sort of measure in place to make sure that future contributors to Litani continue to use this flag/variable when they write new tests. Since we're using Litani to run the test suite, would it be a good idea to add `--timeout 10` and maybe `--timeout-ignore` to `litani add-job` when someone specifies `--fast`?

test/run
```diff
@@ -63,6 +86,9 @@ def add_e2e_tests(litani, test_dir, root_dir, counter):
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
 
+        if fast and get_test_module(test_file).SLOW:
```
karkhaz: Please check `SLOW` has been defined in the module. It is an error if the variable does not exist.

karkhaz (changes requested):

karkhaz:
> I think it would be good to have some sort of measure in place to make sure that future contributors to Litani continue to use this flag/variable when they write new tests

agree, the test script should check that `SLOW` has been defined (either True or False) in every test module and fail if not.

> would it be a good idea to add --timeout 10 and maybe --timeout-ignore to litani add-job when someone specifies --fast?

This is a great idea. No need for `--timeout-ignore` I think, the test should just fail if it times out IMO.

test/README
```diff
@@ -4,3 +4,9 @@
 The `run` script uses litani to run a variety of unit and system tests.
 The dashboard showing the output of the latest test run will always be
 symlinked from test/output/latest/html/index.html.
+
+# OPTIONS
+
+*--fast*
+	Do not run e2e tests with SLOW variable set to true
+    If the run time of a test is >=10, SLOW=True else SLOW=False
```
angelonakos: > If the run time of a test is >=10, SLOW=True else SLOW=False

is missing the unit for time. I would state the following:
```
If a test does not contain a boolean variable called `SLOW` that e2e test will 
not run even if this option is specified. If the run time of an e2e test exceeds 10
seconds, then the value for the constant called SLOW should be set to True
otherwise it should be set to False.
```

nrdg42: Is the description of the flag the best place to put the description of the variable? Additionally is it only in the e2e tests that we need this flag to be set or should it be set every tests? I would prefer

```
## OPTIONS

*--fast*
    Do not run tests with SLOW variable set to True

## Writing Tests

Each test must set a variable called SLOW.
If the run time of a test is >= 10 seconds, SLOW should be set to True, and False otherwise
```

angelonakos: I favor Nick's version as it's more laconic.

ronakfof: Resolved

test/run
```diff
@@ -182,7 +199,7 @@ def main():
 
     add_unit_tests(litani, test_dir, root, counter)
     add_e2e_tests(
-        litani, test_dir, counter, output_dir=output_dir)
+        litani, test_dir, counter, output_dir=output_dir, fast=args.fast)
```
angelonakos: This line could become:

`litani, test_dir, counter, output_dir, args.fast)`

ronakfof: Resolved

test/run
```diff
@@ -79,6 +89,13 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
 
+        try:
```
angelonakos: I propose you change this to:
```python
        if fast and is_slow_test(test_file):
            continue
```

angelonakos: I propose you insert:

```python
sys.path.insert(1, str(e2e_test_dir / "tests")) 
```

before the `for` loop

test/run
```diff
@@ -48,6 +53,11 @@ def run_cmd(cmd):
         sys.exit(1)
 
 
+def get_test_module(module_file):
+    sys.path.insert(1, str(module_file.parent))
```
angelonakos: I propose you change this function to:

```
def is_slow_test(module_file):
    try:
        return importlib.import_module(str(module_file.stem)).SLOW
    except AttributeError:
        logging.error("Variable SLOW is missing from: %s", module_file.name)
        sys.exit(1)
```

angelonakos: Currently, you are inserting `str(module_file.parent))` for each `module_file`.

This results into `/my-custom-path/aws-build-accumulator/test/e2e/tests` being added 21 times (the total number of tests minus 1) being added to `sys.path`. It should be added only once.

angelonakos (changes requested):

angelonakos (approved):

test/README
```diff
@@ -9,3 +9,11 @@ symlinked from test/output/latest/html/index.html.
...
 *--output-dir*
 	Litani will write all of its test output files for this run to this path
+
+*--fast*
+    Do not run tests with SLOW variable set to True
+
+## Writing Tests
+
+Each test must set a variable called SLOW.
+If the run time of a test is >= 10 seconds, SLOW should be set to True, and False otherwise
```
angelonakos: Should this be updated to reflect what you have mentioned in the PR's description?

ronakfof: Resolved

test/run
```diff
@@ -93,7 +109,8 @@ def add_e2e_tests(litani, test_dir, counter, output_dir):
             ci_stage="test",
             description=f"{test_file.stem}: init",
             outputs=run_dir / ".litani_cache_dir",
-            cwd=run_dir)
+            cwd=run_dir,
+            timeout=2 if fast else 0)
```
angelonakos: I'm noticing that you setting `timeout=N if fast else 0` (where `N` is a value) for the different types of Litani jobs.

How did you determine `N`?

Does `timeout=0` cause the Litani job to not time out?

The `--timeout` section in this [page](https://awslabs.github.io/aws-build-accumulator/#litani-man_litani-add-job) does not provide a clean answer to my question.

If `timeout=0` does indeed cause the job to not time out, can we cut a short documentation issue to update this [section](https://github.com/awslabs/aws-build-accumulator/blob/release/doc/src/man/litani-add-job.scdoc#L159-L164)?

ronakfof: When i tested this, `timeout=0` acted as `timeout=None` which was default.
I determined `N` by testing timeouts on jobs which i marked as `fast` and tested multiple times to come to a conclusion that the `10 jobs` with `SLOW=False` follow these timeouts.

nrdg42: I'm not a big fan of repeating the `timeout=2 if fast else 0` logic so many times, can we set `test_timeout=10 if fast else None` and use that variable instead? I feel as though this lines up better with the way that the flag is documented. That way if I were to add a job that takes 9 seconds I would still see it pass.

test/run
```diff
@@ -160,7 +180,7 @@ def print_counter(counter):
 
 def main():
     args = get_args()
-    logging.basicConfig(format="run-tests: %(message)s")
+    logging.basicConfig(format="\nrun-tests: %(message)s")
```
nrdg42, line 211: Can I ask why you're adding a newline here?

test/run
nrdg42: Oh I'm noticing that the documentation was updated to reflect the current timeout of those tests. I'm curious as to why we're being so strict about the timeout on them as they stand now. If we want to add more tests in the future do we want to make it so that they also have to maintain that same pattern of timeouts?

test/README
```diff
@@ -9,3 +9,18 @@ symlinked from test/output/latest/html/index.html.
...
+
+Each test must set a variable called SLOW.
+We have 4 jobs per test (init, add-jobs, run-build, check-run) -
+If run time for all these 4 jobs for a test is <= 14s, then SLOW=False for that test else SLOW=True
+
+Added timeouts for the mentioned jobs is --fast is passed -
+    init - 2s
+    add-jobs - 2s
+    run-build - 8s
+    check-run - 2s
```
nrdg42: > If run time for all these 4 jobs for a test is <= 14s, then SLOW=False for that test else SLOW=True

Not necessarily true with the breakdown of the timeouts. If add-jobs takes 3 seconds but run-build only takes 5, the runtime is <=14 but the jobs would still time out. 

```
Each test is made up of 4 jobs: init, add-jobs, run-build, and check-run
If SLOW is set, the jobs are given the following timeouts
    init - 2 seconds
    add-jobs - 2 seconds
    run-build - 8 seconds
    check-run - 2 seconds
```

test/run
ronakfof: In case of this error or basically any logging statements in code - 

With `\n`
```
17 / 93 tests added
run-tests: Variable SLOW is missing from: custom_stages.py
```

Without `\n`
```
17 / 93 tests addedrun-tests: Variable SLOW is missing from: custom_stages.py
```

test/README
ronakfof: Resolved

test/run
ronakfof: I updated all timeouts to be 10s like you suggested.
It is totally up to us to be strict and pass timeout as well with jobs to be sure that for any test which takes longer than the timeout, SLOW is set to True

test/README
```diff
@@ -9,3 +9,12 @@ symlinked from test/output/latest/html/index.html.
...
 *--output-dir*
 	Litani will write all of its test output files for this run to this path
+
+*--fast*
+    Do not run tests with SLOW variable set to True
+
+## Writing Tests
+
+Each test is made up of 4 jobs: init, add-jobs, run-build, and check-run
+If --fast flag is passed, the jobs are given the a timeout=10s
```
angelonakos: >  If --fast flag is passed, the jobs are given the a timeout=10s

I suggest: ` If --fast flag is passed, each job is given a timeout of 10 seconds. In the worst case, a test would take 40 seconds to run.`

test/README
```diff
@@ -9,3 +9,12 @@ symlinked from test/output/latest/html/index.html.
...
+For each test, SLOW is set to True is jobs fail to pass the above timeout.
```
angelonakos: >  For each test, SLOW is set to True is jobs fail to pass the above timeout.

Can you rephrase?

karkhaz (approved):

karkhaz:
Please resolve conflicts manually, amend your commit, force-push, and merge. It looks like an easy change, you want to keep both my code and your code.

nrdg42 (approved):

Issue #84, opened 2022-01-21
Do not print error messages from tput when not connected to a tty


karkhaz:
From @celinval. When Litani is not connected to a terminal (e.g. in CI or as a background process), it repeatedly prints the message
```
tput: No value for $TERM and no -T specified
```
I believe this text comes from the stderr of `tput` which we run from `ninja.py`.

## Acceptance
If Litani is not connected to a terminal, it should not print that message. Probably just set `stderr=subprocess.DEVNULL`, not sure if something else is required.

celinval:
Is it possible to avoid calling tput?

celinval:
It would be great if this also fixes our CI output. I noticed that the output gets quite messy when $TERM is not set, this is what I get with a valid $TERM set: 

```
[1908/1916] Ca Kani reason about it?
FAILED: /home/ubuntu/rmc/build/output/litani/runs/83c6d59a-3f9f-44e6-abde-73053af6bfcc/status/4e2ad8ff-db74-4ea5-8abe-352a377ea177.json 

/home/ubuntu/rmc/src/tools/litani/litani exec --command 'RUSTFLAGS="--edition 2015" "kani" "src/test/bookrunner/books/Rust by Example/Error handling/Iterating over Results/54.rs"' --pipeline-name 'src/test/bookrunner/books/Rust by Example/Error handling/Iterating over Results/54.rs' --ci-stage verification --job-id 4e2ad8ff-db74-4ea5-8abe-352a377ea177 --description 'Ca Kani reason about it?' --timeout 60 --status-file /home/ubuntu/rmc/build/output/litani/runs/83c6d59a-3f9f-44e6-abde-73053af6bfcc/status/4e2ad8ff-db74-4ea5-8abe-352a377ea177.json --profile-memory-interval 10 --inputs 'src/test/bookrunner/books/Rust by Example/Error handling/Iterating over Results/54.codegen' --ok-returns 0

[1909/1916] Ca Kani reason about it? 
FAILED: /home/ubuntu/rmc/build/output/litani/runs/83c6d59a-3f9f-44e6-abde-73053af6bfcc/status/2eab21f5-19c0-4d2e-b37d-e2d9a259b7b0.json 

[...]

PR #85, opened 2022-01-24
Do not print error messages from tput when not connected to a tty

angelonakos:
*Issue #, if available:* #84

*Description of changes:* Litani calls `tput cols` in order to determine the appropriate column width. When the `TERM` environment variable is not set, which is the case for workflows running on Linux-based platforms on continuous integration services like AWS CodeBuild or GitHub actions, then `tput cols` will fail with a message of `tput: No value for $TERM and no -T specified`. In order to avert that error, this change will provide the return-value of `80`.

### DIFF
--- a/lib/ninja.py
+++ b/lib/ninja.py
@@ -57,6 +57,8 @@ class _OutputAccumulator:
 
     @staticmethod
     def get_tty_width():
+        if os.getenv("TERM") is None:
+            return 80
         try:
             proc = subprocess.run(
                 ["tput", "cols"], text=True, stdout=subprocess.PIPE, check=True)


ronakfof (approved):

lib/ninja.py
```diff
@@ -57,9 +57,12 @@ class _OutputAccumulator:
 
     @staticmethod
     def get_tty_width():
+        tput_cmd = ["tput", "cols"]
+        if os.getenv("TERM") is None:
+            tput_cmd.extend(["-T", "linux"])
```
karkhaz: I'm not sure if this is a good idea. This is going to look up the terminal width in the 'linux' terminfo settings, but I don't see why this would give us an accurate number if `TERM` and `COLUMNS` aren't defined. For example in CI, it's not at all clear to me that whatever the '`linux`' entry is will correspond to the actual width of the CI logging system.

My suggestion is: run the command, and if it fails, return a safe and sensible default like 70.

karkhaz (changes requested):

lib/ninja.py
nrdg42: Just tried this on my machine after looking at the man pages for tput and didn't understand how this would fix the issue. Here's what happened when I ran it on my machine

image

angelonakos: For Ubuntu environments in both AWS CodeBuild and GitHub Actions the`tput cols -T linux` command will return 80.

angelonakos: I created a workflow in GitHub Actions to test this out ([link](https://github.com/angelonakos/aws-build-accumulator/runs/4928326871?check_suite_focus=true)).

karkhaz: Right, because `linux` is not in your terminfo database---not surprising if you're on macOS.

(To find the list of terminal types on my system, I ran `infocmp -D` which printed the path to the database (`/usr/share/terminfo` on my system), then I ran `find /usr/share/terminfo`. That's a big list of terminal types but it will certainly vary from one system to another.)

I agree, using a random value for `TERM` doesn't seem robust, unless there's some terminal type that's guaranteed to exist on all systems... I'm not sure that there is.

karkhaz: Right, but as noted below, this is not robust for systems that don't know what a `linux` terminal is, including macOS apparently...

angelonakos: I also tested in an AWS CodeBuild project that ran in a Linux environment type. 

Specifically, if one takes the below code, places it in a directory called `foo`, compresses it  into a .zip, uploads it to an S3 bucket, creates an AWS CodeBuild project running on Linux, whose buildspec takes the single command of `python3 foo/foo.py` , and finally attempts to start a build, they will notice that the build succeeds and AWS CodeBuild printing out in the build logs `there was an output of 80`. This the code that I mentioned:


```python
#!/usr/bin/env python3

import logging
import os
import subprocess


def main():
    tput_cmd = ["tput", "cols"]
    if os.getenv('TERM') is None:
        tput_cmd.extend(["-T", "linux"])
    try:
        proc = subprocess.run(
            tput_cmd, text=True, stdout=subprocess.PIPE, check=True)
    except (FileNotFoundError, subprocess.CalledProcessError):
        return None
    if proc.returncode:
        logging.info("there was a return code of %s", proc.returncode)
        return None
    try:
        _output = proc.stdout.strip()
        logging.info("there was an output of %s", _output)
        return int(_output)
    except TypeError:
        return None


if __name__ == "__main__":
    logging.basicConfig(format="%(message)s", level=logging.INFO)
    main()

```

angelonakos: Your system has `TERM` set. The control flow of `get_tty_width()` will not append `-T linux` to the `tput_cmd` variables. The function would return 83 in your case.

angelonakos: As I mentioned in the description, the value for `TERM` is not set when running in Linux-based environments in AWS CodeBuild and GitHub actions. Adding the `-T linux` in those cases will result into the error message `tput: No value for $TERM and no -T specified` no longer being printed. A column width of 80 will be used in those cases.

angelonakos: I think the easiest solution is to return 80 in the case that `TERM` is not set and forgo changing the "`tput` command" to include `-T linux`.

Also, I tried changing the `stdout` to `subprocess.DEVNULL`, but that didn't work for me.

angelonakos:
The column width will be set to 80 in the case that `TERM` is not set.

lib/ninja.py
angelonakos: > Right, because `linux` is not in your terminfo database---not surprising if you're on macOS.

In a new zsh terminal:

```bash
> echo $TERM
xterm-256color
> tput cols -T xterm-256-color
80
> tput cols -T linux
80
> unset TERM
> tput cols -T xterm-256-color
tput: No value for $TERM and no -T specified
> tput cols -T linux
tput: No value for $TERM and no -T specified
```

The error is not because `linux` is not in the terminfo database, but because there is no value for `$TERM`.

karkhaz (approved):

karkhaz:
Thanks, let's merge this!

I opened #86 to address the broader wastefullness of the terminal width code, but that's an optimization and not as urgent.

nrdg42 (approved):

PR #89, opened 2022-01-27
Change man page extension from .roff to its chapter

ronakfof:
*Issue #, if available:*

*Description of changes:*
Updated `doc/configure` script to render `scdoc` file to `.1` pages instead of `.roff` in order to use `man litani` or `man litani -init` commands after installation of Litani using brew.

### DIFF
--- a/doc/configure
+++ b/doc/configure
@@ -17,8 +17,6 @@
 import sys
 import pathlib
 
-import jinja2
-
 
 DOC_DIR = pathlib.Path(sys.path[0]).resolve()
 sys.path.insert(1, str(DOC_DIR.parent))
@@ -36,19 +34,19 @@ SRC_DIR = DOC_DIR / "src"
 TEMPLATE_DIR = DOC_DIR / "templates"
 TMP_DIR = DOC_DIR / "tmp"
 
-ROFF_DIR = OUT_DIR / "man"
+MAN_DIR = OUT_DIR / "man"
 
-HTML_MAN_SRC_DIR = TMP_DIR / "roff_to_html"
+HTML_MAN_SRC_DIR = TMP_DIR / "man_to_html"
 HTML_UNIQUE_DIR = TMP_DIR / "html_unique"
 
 
 RULES = [{
-    "name": "sc_to_roff",
-    "description": "converting ${man-name} to roff",
+    "name": "sc_to_man",
+    "description": "converting ${man-name} to man",
     "command": "scdoc < ${in} > ${out}"
 }, {
-    "name": "voluptuous_to_roff",
-    "description": "converting ${man-name} to roff",
+    "name": "voluptuous_to_man",
+    "description": "converting ${man-name} to man",
     "command": f"{BIN_DIR / 'schema-to-scdoc'}"
                f" --project-root { DOC_DIR.parent }"
                " --page-name ${man-name}"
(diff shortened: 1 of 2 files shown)


karkhaz:
I don't think all the man pages should go in `man1`. Each `manx` directory contains man pages in Chapter `x`. You'll find that some litani man pages are in chapter 1, others are in chapter 5 or 7. You need to read the first line of the `.scdoc` file to find the chapter number.

So this needs to be changed so that the configure script figures out, from each `.scdoc` page, which output directory the generated man page should be written to. One way is to read the first line of the file. Another probably easier option is to rename the files, e.g. from `litani-add-job.scdoc` to `litani-add-job.1.scdoc`, but I don't like this because it duplicates the information that's already inside the file.

karkhaz:
So just to be clear: all man pages that go into `man5` need to be saved with a `.5` extension instead of `.roff`, also.

ronakfof:
> So just to be clear: all man pages that go into `man5` need to be saved with a `.5` extension instead of `.roff`, also.

What about `litani-run.json.yaml` and `litani-outcome-table.json.yaml` in `doc/src/voluptuous-man`? What should be their extension and where should they go?

karkhaz:
Those are both in chapter 5. You're right, I just noticed that the chapter number isn't in the source code, but everything under voluptuous-man goes in chapter 5.

You can see the chapter numbers in the HTML documentation on the left: https://awslabs.github.io/aws-build-accumulator/

doc/configure
```diff
@@ -111,30 +109,36 @@ def roff_to_html(man, roff_out, builds):
     return html_man
 
 
-def convert_man_dir_to_roff(src_dir, dst_dir, rule, html_mans, builds):
+def convert_man_dir_to_man(src_dir, dst_dir, rule, html_mans, builds):
     for man in (src_dir).iterdir():
-        roff_out = dst_dir / f"{man.stem}.roff"
+        with open(man) as fp:
+            line = fp.readline().rstrip()
+        index = line[line.find('(')+1:line.find(')')]
```
karkhaz: I don't think this works for the YAML files under voluptuous-man. Those files don't have a first line with a chapter number.

I think the three lines above should be moved to inside the `if` statement below. They are only applicable to `.scdoc` files.

doc/configure
```diff
@@ -57,8 +55,8 @@ RULES = [{
                f" { TEMPLATE_DIR / 'voluptuous-man.jinja.scdoc' }"
                " | scdoc > ${out}"
 }, {
-    "name": "roff_to_html",
-    "description": "converting ${man-name}.roff HTML",
+    "name": "man_to_html",
+    "description": "converting ${man-name}.1 HTML",
```
karkhaz: This isn't always going to be `.1`. I suggest to leave the extension out completely, just use `${man-name}`. This is just debugging output so it doesn't really matter

karkhaz (changes requested):

karkhaz:
Please could you also fill in a complete commit message? In the title, use "its" rather than "it's". The body should mention that this is in preparation for Litani to be installed on users' systems, where the man pages need to be installed in the correct location and have the correct prefix in order for the `man` program to find them.

doc/configure
ronakfof: Resolved

doc/configure
ronakfof: Resolved

karkhaz (approved):

angelonakos (approved):

PR #90, opened 2022-01-28
Render rich output in dashboard, pipeline pages

### LINKED ISSUE #81: Allow jobs to emit arbitrary text to be inlined onto pipeline page
Currently, the pipeline page displays a box for each job with some fixed information: stdout/stderr, return codes, etc.

Users sometimes want to display richer data than  just the text of stdout/stderr. For example, the result of a job might be more nicely represented as a graph or as rendered HTML. Users would thus want a way to make a job print an SVG on its stdout, and a way of telling Litani that the stdout should be inlined on to the web page rather than being treated as text.

Acceptance:
* The Litani report renderer changes its behaviour if users gave a special tag to a job, like `litani add-job --tags literal-stdout`. If a job has that tag, instead of displaying a black box with stdout/stderr inside an `<xmp>` tag, the stdout gets inlined onto the page literally inside the div. The user is expected to print out an SVG or fragment of HTML that works with the rest of the page.

### LINKED ISSUE #82: Allow jobs to emit arbitrary HTML to be rendered onto front page
Similar to #81. Users want to have Litani jobs that emit HTML or SVG that should be inlined directly onto the front page.

Acceptance:
* If any job is tagged with some tag, e.g. `front-page-text`, then the Litani front page will include a new section called "Job Outputs". For each job that emits verbatim text, that text should be inlined directly onto the front page, maybe with a title/hyperlink to indicate which job it came from. So like this:

# s2n proof results
```
... progress bars ...
```
### Output of job ["Overall coverage"](/):
```
... the literal stdout of that job. It's expected to be well-formed HTML, SVG etc
```
### Output of job ["Sizes of SAT formulas"](/):
```
... another graph
```

## Parallelism
```
...
```

angelonakos:
This feature allows users to render custom HTML data onto the HTML dashboard, allowing Litani jobs to display their results through tables, graphs, and other HTML format. This can be done both for the front page (example [here](https://github.com/awslabs/aws-build-accumulator/pull/90#issuecomment-1029432735)) or on the pipeline page ([example](https://github.com/awslabs/aws-build-accumulator/pull/90#issuecomment-1027979021)).

The intention is that individual litani jobs can be used to measure metrics, and then print those metrics out in an easily-viewable format. Jobs that calculate metrics for the entire run can present those metrics as a graph on the front page. Jobs that calculate metrics for a single pipeline (or proof) can display the result on the pipeline page.

Users use this feature by adding a tag to a litani job, using the `--tags` flag. If a job is tagged with `front-page-text`, Litani will render the job's output onto the front page, in its own section. A tag of `literal-stdout` will make Litani render the job's output onto the pipeline page, but without any HTML escaping.

This feature fixes #81 and fixes #82.


templates/pipeline.jinja.html
```diff
@@ -773,12 +773,21 @@ <h1>
...
-            </xmp>
-          </div><!-- class="output-box" -->
+            {% if job["wrapper_arguments"]["tags"] is not none and "literal-stdout" in job["wrapper_arguments"]["tags"] %}
+            <div class="pseudo">
+              {% for line in job["stdout"] %}
+  {{ line.strip() }}
+              {%- endfor %}{# line in job["stdout"] #}
+            </div><!-- class="pseudo" -->
+            {% endif %}{# literal-stdout is in tags #}
+            {% if job["wrapper_arguments"]["tags"] is none %}
```
karkhaz: I don't think this is correct. There are other unrelated tags that we already use, e.g. `stats-group:xxx` (have a look in the starter kit to see what that's for).

I think it should just be `else`.

templates/pipeline.jinja.html
```diff
@@ -773,12 +773,21 @@ <h1>
...
 
           {% if job["stdout"] %}
-          <div class="output-box"><xmp>stdout:
-            {% for line in job["stdout"] %}
-{{ line.strip() }}
-            {%- endfor %}{# line in job["stdout"] #}
-            </xmp>
-          </div><!-- class="output-box" -->
+            {% if job["wrapper_arguments"]["tags"] is not none and "literal-stdout" in job["wrapper_arguments"]["tags"] %}
+            <div class="pseudo">
```
karkhaz: what is `pseudo`? You haven't added any particular style for it?

karkhaz (changes requested):
Thanks, looking forward to this!

This is a pretty important feature, I'd ideally like to be able to just paste the commit message into the changelog for next release. Could you elaborate on why this change is useful/needed? I mentioned this briefly in #81 but I think it's good for it to be duplicated/elaborated in the commit history, since that's what we generate changelogs from.

templates/pipeline.jinja.html
nrdg42: `pseudo` sounds like it would be a wrapper to group whatever elements the user wants to render. Maybe the div should be given an `id=literal-stdout` tag?

karkhaz: There could be one of this div per job on the pipeline page, so we can't give it an `id` but I agree with `class=literal-stdout`.

But I understand, now, of course there's no style because there isn't really any style that we could reasonably give to it. The stdout is expected to provide its own styling.

Perhaps, if anything, the div could have a special border to indicate that everything inside it is program output rather than something that litani generated directly? Like a border with a header at the top that says "literal stdout"? Though without a border it probably looks cleaner.

angelonakos:
I copied the style of the "output box" class for this "literal stdout" class. Also, I ran the following:

```bash
python3 litani init --project-name issue-81 && \
python3 litani add-job --command 'curl zappos.com' --pipeline-name 'my-cool-pipeline' --ci-stage build --phony-outputs a-phony-output && \
python3 litani add-job --command "echo '<img src=\"https://raw.githubusercontent.com/aws/s2n-tls/main/docs/images/s2n-logo-final.svg\" alt=\"s2n logo\"/>'" --pipeline-name 'my-cool-pipeline' --ci-stage build --inputs a-phony-output --phony-outputs abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz --tags literal-stdout  && \
python3 litani add-job --command 'for i in 1 2 3 4 5; do dd bs=1 count=8 if=/dev/urandom 2>/dev/null|base32|head -c 8; done' --pipeline-name 'my-cool-pipeline' --ci-stage test --inputs abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz --phony-outputs 123 --tags literal-stdout && \
python3 litani add-job --command "curl -L amazon.com" --pipeline-name 'my-cool-pipeline' --ci-stage report --inputs 123 --phony-outputs 456 --tags literal-stdout && \
python3 litani run-build
Report will be rendered at file:///var/folders/bd/w7pff05n5k35735rjvph7dkh0000gs/T/litani/runs/latest/html/index.html
[4/4] /Volumes/workplace/bt101/aws-build-accumulator/litani exec --command 'curl -L amazon.com' --pipeline-name my-cool-pipeline --ci-stage report --job-id cc13d395-4204-4374-a7b5-92918ad3209a --...
Report was rendered at file:///var/folders/bd/w7pff05n5k35735rjvph7dkh0000gs/T/litani/runs/latest/html/index.html
```

This will render text within the usual stdout, stderr boxes.

When the `literal-stdout` tag is provided:

- An SVG will be rendered both within the "terminal box" and the "output box".
- Text will be rendered within the "output box".
- HTML will be rendered within the "output box".

angelonakos:
>    An SVG will be rendered both within the "terminal box" and the "output box".

Is the fact that an SVG can be rendered both in the "command content box" and inside of the "output box" an issue? Should SVGs not be rendered in the "command content box"?

angelonakos:
image

karkhaz:
This looks terrific. The green box shouldn't be used for displaying the SVG, it's only ever for printing the command, I'm not completely sure why the SVG got displayed there.

As discussed offline, please could you add some more realistic examples. I'm also interested in comments from others about whether/how they would use this; whether it's easy to use (is 'literal-stdout' an okay name?), etc. I also wonder whether it would be worth addressing #82 either as a second commit in this PR or a separate PR, again to get comments from users.

nrdg42:
I've just added commits to address issue #82, as well as examples showing the `front-page-text` and the `literal-stdout` tags.
I've included screenshots of `run-2.sh` and `run-3.sh` from `examples/rich-output` in both light and dark mode. I was unable to get `run-1.sh` to generate HTML consistently, and I'd appreciate any insight on that. 
image
image
image
image

karkhaz:
I don't think I was clear in my comment, but I wasn't expecting you to remove the yellow and green boxes.

However... maybe it's actually cleaner to remove them? I think it's still useful to have them to see what command was used to display that output. But also curious to hear what others think.

doc/src/man/litani-add-job.scdoc
```diff
@@ -123,7 +123,10 @@ configure times.
 	dashboard generator does use some of them). Each tag can be in whatever format
 	you prefer, e.g. a plain string, key-value pair, or even arbitrary JSON. A
 	job's list of tags is included in its _run.json_ and is intended to help with
-	analyzing run data.
+	analyzing run data. If `literal-stdout` is provided as a tag, then the
+	report will render the literal standard output generated by the job. If
+	'front-page-output' is provided as a tag, then the report will render the
```
angelonakos: We should change this to `front-page-text`

examples/rich-output/README.md
```diff
@@ -0,0 +1,10 @@
+README
+
+The run-all.py script will run examples that demonstrate how jobs in Litani can
+display "rich" output either on the dasboard or pipeline pages
```
adpaco:
```suggestion
display rich output either on the dashboard or pipeline pages
```

doc/src/man/litani-add-job.scdoc
```diff
@@ -123,7 +123,10 @@ configure times.
 	dashboard generator does use some of them). Each tag can be in whatever format
 	you prefer, e.g. a plain string, key-value pair, or even arbitrary JSON. A
 	job's list of tags is included in its _run.json_ and is intended to help with
-	analyzing run data.
+	analyzing run data. If `literal-stdout` is provided as a tag, then the
```
adpaco: Reading this it is not clear to me what you mean by "literal standard output" or "literal standard out" (which seem to be different concepts).

If this feature allows you to display images and graphs, just say so here and maybe link to more precise documentation regarding the feature (e.g., what are the various outputs you can display or formats you support, where to find examples, etc.).

examples/rich-output/assumptions.html
```diff
@@ -0,0 +1,41 @@
...
+                    <td>Medium</td>
+                </tr>
+                <tr>
+                    <td>Function specifications</td>
+                    <td>Hard</td>
+                </tr>
+            </table>
+        </div>
+    </body>
+</html>
```
adpaco: Add a newline.

examples/rich-output/file.dat
```diff
@@ -0,0 +1,7 @@
+pipeline	#Assumptions
+"Class\\\_A"	5
+"Class\\\_B"	1
+"Class\\\_C"	2
+"Class\\\_D"	8
+"Class\\\_E"	4
+"Class\\\_F"    0
```
adpaco: Adjust spacing here.

examples/rich-output/scripts/fib.py
```diff
@@ -0,0 +1,40 @@
...
+    for i in range(10):
+        if i == 0:
+            fib.append(a)
+        elif i == 1:
+            fib.append(b)
+        else:
+            fib.append(a + b)
+            temp = b
+            b = a + b
+            a = temp
```
adpaco: A simpler way of computing fibonacci numbers is:
 1. Allocate a list of zeros with length equal to `n` (i.e., `[0]` * 10)
 2. Assign 1 to the second element (`fib[1] = 1`)
 3. Iterate from 2 to `n` doing `fib[i] = fib[i-1] + fib[i-2]`

examples/rich-output/scripts/sin.py
```diff
@@ -0,0 +1,30 @@
...
+import math
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--output", required=True, help="output file")
+    args = parser.parse_args()
+    with open(args.output, "w") as handle:
+        for i in range(9):
+            val = ((2 * math.pi) / 8) * i
```
adpaco: Would use floating point constants and castings here.

templates/pipeline.jinja.html
```diff
@@ -773,12 +781,20 @@ <h1>
...
-          <div class="output-box"><xmp>stdout:
-            {% for line in job["stdout"] %}
-{{ line.strip() }}
-            {%- endfor %}{# line in job["stdout"] #}
-            </xmp>
-          </div><!-- class="output-box" -->
+            {% if job["wrapper_arguments"]["tags"] is not none and "literal-stdout" in job["wrapper_arguments"]["tags"] %}
+            <div class="literal-stdout">
+              {% for line in job["stdout"] %}
+  {{ line.strip() }}
```
adpaco: Can this be indented or would it affect the output?

nrdg42:
> I don't think I was clear in my comment, but I wasn't expecting you to remove the yellow and green boxes.
> 
> However... maybe it's actually cleaner to remove them? I think it's still useful to have them to see what command was used to display that output. But also curious to hear what others think.

@karkhaz If you're referring to the screenshots I posted, those are on Litani's dashboard using a new tag which is different from Angelo's `literal-stdout`. The title of the diagram links to the job with the yellow and green boxes on the pipeline page. Do you want the yellow and green boxes on the dashboard as well?

karkhaz:
Oh I see, yes, your screenshots are from the front page. Never mind :) They look great!

doc/src/man/litani-add-job.scdoc
angelonakos: I rephrased these sentences to eliminate confusion wrt. "literal stdout" and to elaborate on the exact capabilities with references to examples.

examples/rich-output/scripts/fib.py
angelonakos: More efficient. Thanks. Changed accordingly.

examples/rich-output/scripts/sin.py
angelonakos: Made changes to both use floating point constants and eliminated rounding from the underneath line.
image

templates/pipeline.jinja.html
angelonakos: Indentations affects the output. This change will have to remain this way.

adpaco-aws (approved):

ronakfof (approved):

lib/litani_report.py
```diff
@@ -570,11 +562,27 @@ def render(run, report_dir, pipeline_depgraph_renderer):
                     MemoryTraceRenderer.render(
                         temporary_report_dir / pipe["url"], env, job, gnuplot)
 
+                tags = job["wrapper_arguments"]["tags"]
+                description = job["wrapper_arguments"]["description"]
+                if tags and "front-page-text" in tags:
+                    if "stdout" in job:
```
karkhaz: I think `if "stdout" in job and job["stdout"]`, since even if the key exists, its value may be None as per the schema.

templates/dashboard.jinja.html
```diff
@@ -445,6 +490,30 @@ <h1>
...
+      <div class="job-outputs-header">
+        <h2>Job Outputs</h2>
+      </div>
+      {% for job_desc, job in front_page_outputs.items() %}
+        <div class="job-output-content">
+          <div class="job-output-title">
+            <h3>Output of job
+              <a href="pipelines/{{ job['wrapper_arguments']['pipeline_name'] }}
+/index.html#job-{{ job['wrapper_arguments']['job_id'] }}">
+                "{{ job_desc }}"
```
karkhaz: Nit: quotes should be outside `<a>`. Also, use `&quot;`.

karkhaz (changes requested):

karkhaz (approved):

Issue #92, opened 2022-02-03
Dependency Graph prints file extensions twice


nrdg42:
While working on Litani I noticed that file extensions on the dependency graph get printed twice when they should just be printed once. In this case `fib.csv.csv` should just be `fib.csv`

![image](https://user-images.githubusercontent.com/68433946/152380355-a774352f-d973-4a0f-85f9-d27a4cd45d00.png)

PR #93, opened 2022-02-03
Prevent file extension from appearing twice in dot dependency diagram

angelonakos:
*Issue #, if available:*

*Description of changes:*

Fixes: #92

### DIFF
--- a/lib/graph.py
+++ b/lib/graph.py
@@ -56,12 +56,12 @@ class DependencyNode(Node):
         self.id = hash(fyle)
 
         path = pathlib.Path(fyle)
-        _, ext = os.path.splitext(path.name)
+        name, ext = os.path.splitext(path.name)
 
         self.style = style
 
         wrapper = get_text_wrapper(line_width)
-        path_name = "\n".join(wrapper.wrap(path.name))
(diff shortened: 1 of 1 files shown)


lib/graph.py
```diff
@@ -56,12 +56,12 @@ def __init__(self, fyle, line_width=40, **style):
...
 
         path = pathlib.Path(fyle)
-        _, ext = os.path.splitext(path.name)
+        name, ext = os.path.splitext(path.name)
 
         self.style = style
 
         wrapper = get_text_wrapper(line_width)
-        path_name = "\n".join(wrapper.wrap(path.name))
+        path_name = "\n".join(wrapper.wrap(name))
```
nrdg42, line 64: I'm confused as to why we wrap the path name without the extension? Wouldn't it make more sense to wrap it all together?

karkhaz: I guess I did this because the wrapper prefers to break lines at spaces and punctuation including periods, and I didn't want it to put the extension on its own line. So we wrap first, and then the extension gets put on the same line as the rest of the file name

karkhaz (approved):

ronakfof (approved):

PR #94, opened 2022-02-03
Add release workflow for litani

ronakfof:
Tests
[Example Release](https://github.com/ronakfof/litani/releases/tag/litani-1.23.4) - [[workflow](https://github.com/ronakfof/litani/runs/5147017859?check_suite_focus=true)]
[Deb package](https://github.com/ronakfof/litani/releases/download/litani-1.23.4/litani-1.23.4.deb) - [[workflow](https://github.com/ronakfof/litani/runs/5147020743?check_suite_focus=true)]
[Example PR to tap](https://github.com/ronakfof/homebrew-litani/pull/11) - [[workflow](https://github.com/ronakfof/litani/runs/5147020632?check_suite_focus=true)]

*Description of changes:*

This workflow gets triggered when a tag is pushed of format `litani-*`.
Bot creates a release for that tag with a boilerplate description.
If a release is created successfully, it also creates a `debian` package for Litani on `ubuntu-20.04`, tests it and uploads it as an additional asset to the release and also create a PR to `aws/tap` with updated `version`, `tag` and `commit`.

### DIFF
--- a/.github/workflows/release-packages.yaml
+++ b/.github/workflows/release-packages.yaml
@@ -1,0 +1,86 @@
+on:
+  release:
+    types: [created]
+
+name: Upload additional release assets
+jobs:
+  ubuntu-20_04-package:
+    name: Generate ubuntu-20.04 debian package
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Version
+        run: echo "VERSION=$(echo ${{ github.ref }} | cut -d "/" -f 3 | cut -d "-" -f 2)" >> $GITHUB_ENV
+      - name: Setup directory for deb package
+        run: |
+          sudo apt-get install -y mandoc scdoc ninja-build
+          mkdir -p litani-${{ env.VERSION }}/{DEBIAN,usr/{bin,libexec/litani,share/{doc/litani,man/{man1,man5,man7}}}}
+          touch litani-${{ env.VERSION }}/DEBIAN/control
+          cat << EOF > litani-${{ env.VERSION }}/DEBIAN/control
+          Package: Litani
+          Version: ${{ env.VERSION }}
+          Architecture: amd64
+          Depends: ninja-build, gnuplot, graphviz, python3-jinja2
+          Maintainer: Kareem Khazem <karkhaz@amazon.co.uk>
+          Description: A program that provides platform-independent job control.
+            AWS Build Accumulator collects build jobs from multiple sources before executing
+            them concurrently. It provides platform-independent job control (timeouts,
+            return code control) and an output format that is easy to render into reports
+            (for example, using the built-in renderer). AWS Build Accumulator shines where
+            your project uses multiple different build systems or requires a unified
+            interface describing heterogeneous build jobs.
+          EOF
+          ./doc/configure && ninja
+          mv bin lib templates litani litani-${{ env.VERSION }}/usr/libexec/litani/
+          mv doc/out/man/*.1 litani-${{ env.VERSION }}/usr/share/man/man1
+          mv doc/out/man/*.5 litani-${{ env.VERSION }}/usr/share/man/man5
+          mv doc/out/man/*.7 litani-${{ env.VERSION }}/usr/share/man/man7
+          mv doc/out/html/index.html litani-${{ env.VERSION }}/usr/share/doc/litani
+          ln -s /usr/libexec/litani/litani litani-${{ env.VERSION }}/usr/bin/
+          rm -r $(ls -A | grep -v litani-${{ env.VERSION }})
+      - name: Create .deb package
+        id: create_packages
+        run: |
+          sudo dpkg-deb --build --root-owner-group litani-${{ env.VERSION }}
+          deb_package_name="$(ls *.deb)"
+          echo "::set-output name=deb_package::$deb_package_name"
+          echo "::set-output name=deb_package_name::$deb_package_name"
+      - name: Install Litani using deb package
+        run: sudo apt-get update && sudo apt install -y ./litani-${{ env.VERSION }}.deb
+      - name: Test deb package
+        run: |
+          litani -h
+          man litani
+          litani init --project-name test
+          litani add-job --command '/usr/bin/true' --pipeline-name 'test' --ci-stage test
+          litani run-build
+      - name: Get release
+        id: get_release_info
+        uses: bruceadams/get-release@v1.2.3
+      - name: Upload release binary
+        uses: actions/upload-release-asset@v1.0.2
+        with:
+          upload_url: ${{ steps.get_release_info.outputs.upload_url }}
+          asset_path: ${{ steps.create_packages.outputs.deb_package }}
+          asset_name: ${{ steps.create_packages.outputs.deb_package_name }}
+          asset_content_type: application/x-deb
+
+  homebrew-pr:
+    name: Homebrew Bump Formula PR
+    runs-on: macos-10.15
+    steps:
+      - name: Get release tag name
+        run: echo "RELEASE_TAG=${GITHUB_REF/refs\/tags\/}" >> $GITHUB_ENV
+      - name: Configure git user name and email
+        uses: Homebrew/actions/git-user-config@master
+        with:
+          username: aws-build-accumulator-release-ci
+      - name: Create homebrew PR
+        run: |
+          brew update-reset
+          brew tap aws/tap
+          brew bump-formula-pr --tag "$RELEASE_TAG" --revision "$GITHUB_SHA" aws/tap/litani
+        env:
+          HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.RELEASE_CI_ACCESS_TOKEN }}
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -1,0 +1,50 @@
+on:
+  push:
+    tags:
+      - 'litani-*'
+
+name: Create Release
+
+jobs:
+  perform-release:
+    name: Perform Release
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.RELEASE_CI_ACCESS_TOKEN }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Get Version
+        run: echo "VERSION=$(echo ${{ github.ref }} | cut -d "/" -f 3 | cut -d "-" -f 2)" >> $GITHUB_ENV
+      - name: Create release
+        uses: actions/create-release@v1
+        with:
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        LITANI_VERSION: ${{ needs.perform-release.outputs.version }}
+      steps:
+        - uses: actions/checkout@v2
+          with:
+            submodules: recursive
+        - name: Fetch dependencies
+          run: sudo apt-get install -y mandoc scdoc ninja-build
+        - name: make dir structure
+          run: mkdir -p litani-${{ env.LITANI_VERSION }}/{DEBIAN,usr/{bin,libexec/litani,share/{doc/{litani},man/{man1,man5,man7}}}}
```
angelonakos: Do you think that the `mkdir -p` statement for:

`usr/{bin,libexec/litani,share/{doc/{litani},man/{man1,man5,man7}}}`

needs to be broken up in the sake of readability?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+          run: |
+            touch litani-${{ env.LITANI_VERSION }}/DEBIAN/control
+            cat << EOF > litani-${{ env.LITANI_VERSION }}/DEBIAN/control
+            Package: Litani
+            Version: ${{ env.LITANI_VERSION }}
+            Architecture: amd64
+            Depends: ninja-build, gnuplot, graphviz, python3.9, python3-jinja2, python3-voluptuous
+            Maintainer: Kareem Khazem <karkhaz@amazon.co.uk>
+            Description: A program that provides platform-independent job control.
+              AWS Build Accumulator collects build jobs from multiple sources before executing them concurrently. It provides platform-independent job control (timeouts, return code control) and an output format that is easy to render into reports (for example, using the built-in renderer). AWS Build Accumulator shines where your project uses multiple different build systems or requires a unified interface describing heterogeneous build jobs.
```
angelonakos: Can this line be wrapped at a column width of 80?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+            mv litani-${{ env.LITANI_VERSION }}/usr/libexec/litani/doc/out/man/*.5 litani-${{ env.LITANI_VERSION }}/usr/share/man/man5
+            mv litani-${{ env.LITANI_VERSION }}/usr/libexec/litani/doc/out/man/*.7 litani-${{ env.LITANI_VERSION }}/usr/share/man/man7
+        - name: Remove unnecessary files
+          run: |
+            rm -rf litani-${{ env.LITANI_VERSION }}/usr/libexec/litani/doc
+            sudo rm -rf /usr/libexec/litani
+        - name: Create .deb package
+          id: create_packages
+          run: |
+            sudo dpkg-deb --build --root-owner-group litani-${{ env.LITANI_VERSION }}
```
angelonakos: are `sudo` and `--root-owner-group` necessary here?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+            echo "::set-output name=deb_package_name::$deb_package_name"
+        - name: Setup deb package
+          run: |
+            sudo apt-get update
+            sudo apt install -y ./litani-${{ env.LITANI_VERSION }}.deb
+        - name: Test deb package
+          run: |
+            litani -h
+            man litani
+            litani init --project-name nyx --stages codegen verify check
```
angelonakos: These 3 stage names can be removed.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+            litani add-job --command 'which brew' --pipeline-name 'siesta' --ci-stage codegen --phony-outputs phony_1
+            litani add-job --command 'which python3' --pipeline-name 'siesta' --ci-stage verify --inputs phony_1 --phony-outputs phony_2
+            litani add-job --command 'which docker' --pipeline-name 'siesta' --ci-stage check --inputs phony_2  --phony-outputs phony_3
```
angelonakos: Is there a possibility `docker` might ever be missing?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,146 @@
...
+            litani run-build
+        - name: Get release
+          id: get_release_info
+          uses: bruceadams/get-release@v1.2.3
```
angelonakos: Is the upload_url  the only piece of information needed from this release? Can it be fetched differently without having to invoke the following GitHub action?

https://github.com/bruceadams/get-release

.github/workflows/release.yaml
```diff
@@ -0,0 +1,131 @@
...
+name: Create Release and package
+
+jobs:
+  get-version-information:
+    name: Get Version Information
+    runs-on: ubuntu-20.04
+    outputs:
+      version: ${{ steps.split-version.outputs._1 }}
+    steps:
+      - uses: jungwinter/split@v1
```
karkhaz: Why not v2?

I guess for each of these externally-authored steps that we use, we should audit them and then stick to a particular version, right? What about security issues?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,131 @@
...
+        uses: actions/checkout@v2
+      - name: Create release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          LITANI_VERSION: ${{ needs.get-version-information.outputs.version }}
+        with:
+          tag_name: litani-${{ env.LITANI_VERSION }}
+          release_name: litani-${{ env.LITANI_VERSION }}
+          body: |
```
karkhaz: Is there any way we can include the changelog entry in here?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,131 @@
...
+        - name: Setup directory for deb package
+          run: |
+            sudo apt-get install -y mandoc scdoc ninja-build
+            mkdir -p litani-${{ env.LITANI_VERSION }}/{DEBIAN,usr/{bin,libexec/litani,share/{doc/litani,man/{man1,man5,man7}}}}
+            touch litani-${{ env.LITANI_VERSION }}/DEBIAN/control
+            cat << EOF > litani-${{ env.LITANI_VERSION }}/DEBIAN/control
+            Package: Litani
+            Version: ${{ env.LITANI_VERSION }}
+            Architecture: amd64
+            Depends: ninja-build, gnuplot, graphviz, python3-jinja2, python3-voluptuous
```
karkhaz: Let's not install voluptuous. I'd like to keep this as optional.

.github/workflows/release.yaml
karkhaz: Same with brew.

I would prefer much more trivial commands like `/usr/bin/true`. And probably there only needs to be one job.

karkhaz (changes requested):
Looking forward to this. My main question is whether we can integrate the changelog into the release notes.

Maybe the right thing to do here is: instead of this workflow triggering on `litani-*` tag, it would trigger on any commit that gets pushed with an updated changelog. In that case, it would take the release notes from the changelog, and the workflow itself would create the release tag.

So from a user perspective, when we want to create a new release, we add to the CHANGELOG and push that change to the `develop` branch, and this workflow would merge the `release` branch into that commit and tag it as the next release.

That might be more complicated though because the workflow itself would have to figure out the release number... But I'm quite keen for the release notes to end up in the tag message.

.github/workflows/release.yaml
ronakfof: The reason i am using `mkdir -p` is to not add unnecessary code to the workflow which will increase the length of code as well as be confusing to understand the directory structure which is way better with this one statement.

.github/workflows/release.yaml
ronakfof: Resolved

.github/workflows/release.yaml
ronakfof: `sudo` is required here.
And the `--root-owner-group` flag makes all deb package content owned by the root user, which is the standard way to go. Without such flag, all files and folders would be owned by the user, which might not exist in the system the deb package would be installed to.

.github/workflows/release.yaml
ronakfof: Resolved

.github/workflows/release.yaml
ronakfof: Resolved

.github/workflows/release.yaml
ronakfof: A standard way is to use a GitHub Action to do so as i have seen while investigating. We can create our own to do so but this one is widely used.

.github/workflows/release.yaml
ronakfof: We use this one for a very simple change and I guess version won't matter here but we can definitely keep it updated.
Also I saw this being used in one of the `cbmc` workflows for the same.

.github/workflows/release.yaml
ronakfof: As discussed in the meetings, we will be triggering this workflow only on tag push to release branch.

.github/workflows/release.yaml
ronakfof: Resolved

.github/workflows/release.yaml
angelonakos: What about:
> I guess for each of these externally-authored steps that we use, we should audit them and then stick to a particular version, right? What about security issues?

.github/workflows/release.yaml
angelonakos: I'm fine with sticking with this widely used step. Thanks

.github/workflows/release-packages.yaml
```diff
@@ -0,0 +1,86 @@
...
+  release:
+    types: [created]
+
+name: Upload additional release assets
+jobs:
+  ubuntu-20_04-package:
+    name: Generate ubuntu-20.04 debian package
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
angelonakos, line 11: Should this be `RELEASE_CI_ACCESS_TOKEN`?

ronakfof: Bot is not involved in creation of the `debian` package sop we don't need it here. We can use the default token here.

.github/workflows/release.yaml
ronakfof: Resolved

.github/workflows/release-packages.yaml
angelonakos: Got it.

angelonakos: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret for those not familiar

angelonakos (approved):

karkhaz (approved):
This is awesome :)

PR #97, opened 2022-02-10
Sort pipelines by status and then name

angelonakos:
*Issue #, if available:*

*Description of changes:*

Name will now be used as the second criterion when sorting pipelines.

The order with which pipelines appear in both the HTML dashboard is the same as
the order with which they appear in the run.json. An e2e test was added to 
ensure that the order is indeed the intended one.

### DIFF
--- a/test/e2e/tests/pipeline_order.py
+++ b/test/e2e/tests/pipeline_order.py
@@ -1,0 +1,53 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+SLOW = True
+EXPECTED_PIPELINE_NAMES = ["zeta", "0", "00", "01", "1", "Beta", "alpha",
+    "beta", "beta1", "beta2", "betaa", "delta", "epsilon", "gamma"]
+
+
+def get_init_args():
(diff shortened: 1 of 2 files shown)


test/e2e/tests/pipeline_order.py
```diff
@@ -0,0 +1,78 @@
...
+    return {
+        "kwargs": {
+            "project": "sorted_pipeline_names",
+        }
+    }
+
+
+def get_jobs():
+    jobs = []
+    for element in (
```
karkhaz: `for ci_stage, pipeline in(`

test/e2e/tests/pipeline_order.py
```diff
@@ -0,0 +1,78 @@
...
+
+def check_run(run):
+    assert_names = []
+    assert_stages = []
+    for index, element in enumerate((
+            "zeta", "0", "00", "01", "1", "Beta", "alpha", "beta", "beta1",
+            "beta2", "betaa", "delta", "epsilon", "gamma")):
+        pipeline = run["pipelines"][index]["name"]
+        assert_names.append(pipeline == element)
+        for idx, stage_obj in enumerate(run["pipelines"][index]["ci_stages"]):
```
karkhaz: The rest of this method (from this line down) seems like it's testing something different, though I'm not completely certain because it's also very complicated... it seems like two tests in one? And I'm not really sure that testing that the CI stages are in the right order is necessary, since you've given all the pipelines unique names, so just checking that the names are sorted should be enough?

I don't understand why this test isn't just something like
```
return [p["name" for p in run["pipelines"]] == [
            "zeta", "0", "00", "01", "1", "Beta", "alpha", "beta", "beta1",
            "beta2", "betaa", "delta", "epsilon", "gamma",
]
```

karkhaz (changes requested):

test/e2e/tests/pipeline_order.py
angelonakos: Indeed this is much simpler. I was unnecessarily complicating this test. Checking the order of names alone will suffice

karkhaz (approved):

ronakfof (approved):

PR #113, opened 2022-03-01
Add --out-file flag to litani-dump-run

### LINKED ISSUE #110: Add `--out-file` flag to `litani-dump-run`

ronakfof:
Fixes #110

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/doc/src/man/litani-dump-run.scdoc
+++ b/doc/src/man/litani-dump-run.scdoc
@@ -22,12 +22,13 @@ litani dump-run - Print the current run as a JSON document to stdout
 # SYNOPSIS
 
 *litani dump-run*
-	\[*--retries* _N_]
+	\[*-r*/*--retries* _N_]
+	\[*-o*/*--out-file* _F_]
 
 
 # DESCRIPTION
 
-This program prints the _run.json_ file for an in-progress Litani run to stdout.
+This program prints the run.json file for a Litani run to stdout or to the specified output file.
 The JSON file's schema is documented in *litani-run.json(5)*.
 
 This program is intended to be used while an invocation of *litani-run-build(1)*
@@ -62,6 +63,8 @@ _"complete"_ field set to *true*.
 	every second up to the limit given in this flag. If _N_ is *0*, this program
 	will retry indefinitely---this is not recommended.
 
+*-o* _F_, *--out-file* _F_
+	Print the run to _F_ instead of stdout
 
 # OUTPUT
 
--- a/lib/run_printer.py
+++ b/lib/run_printer.py
@@ -112,8 +112,12 @@ def get_run_checker():
 
 
 
-def _exit_success(run):
-    print(json.dumps(run, indent=2))
+def _exit_success(run, out_file):
+    if out_file:
+        with lib.litani.atomic_write(out_file) as handle:
+            print(json.dumps(run, indent=2), file=handle)
+    else:
+        print(json.dumps(run, indent=2))
     sys.exit(0)
 
 
@@ -122,7 +126,7 @@ def _exit_error():
     sys.exit(0)
 
 
-def _try_dump_run(cache_dir, pid, sleeper, check_run):
+def _try_dump_run(cache_dir, pid, sleeper, check_run, out_file):
     try:
         os.kill(pid, DUMP_SIGNAL)
     except ProcessLookupError:
@@ -132,7 +136,7 @@ def _try_dump_run(cache_dir, pid, sleeper, check_run):
             with open(latest_run_file) as handle:
                 latest_run = json.load(handle)
             check_run(latest_run)
-            _exit_success(latest_run)
+            _exit_success(latest_run, out_file)
         except (
             FileNotFoundError, json.decoder.JSONDecodeError,
             InconsistentRunError):
@@ -142,7 +146,7 @@ def _try_dump_run(cache_dir, pid, sleeper, check_run):
         with open(cache_dir / _DUMPED_RUN) as handle:
             run = json.load(handle)
         check_run(run)
-        _exit_success(run)
+        _exit_success(run, out_file)
     except (
             FileNotFoundError, json.decoder.JSONDecodeError,
             InconsistentRunError):
@@ -161,13 +165,14 @@ async def dump_run(args):
 
     sleeper = BackoffSleeper(jitter=random.random())
     check_run = get_run_checker()
+    out_file = args.out_file if args.out_file else None
 
     if args.retries:
         for _ in range(args.retries):
-            _try_dump_run(cache_dir, pid, sleeper, check_run)
+            _try_dump_run(cache_dir, pid, sleeper, check_run, out_file)
     else:
         while True:
-            _try_dump_run(cache_dir, pid, sleeper, check_run)
+            _try_dump_run(cache_dir, pid, sleeper, check_run, out_file)
     _exit_error()
 
 
--- a/litani
+++ b/litani
@@ -260,6 +260,11 @@ def get_args():
             "help":
                 "how many times to retry loading the run in 1 second intervals"
                 " (Default: %(default)s)"
+        }, {
+            "flags": ["-o", "--out-file"],
+            "metavar": "F",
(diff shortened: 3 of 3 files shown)


lib/run_printer.py
```diff
@@ -112,8 +112,12 @@ def get_run_checker():
 
 
 
-def _exit_success(run):
-    print(json.dumps(run, indent=2))
+def _exit_success(run, out_file):
+    if out_file:
+        with open(out_file, mode="w") as output_file:
```
karkhaz: Please use `lib.atomic_write` instead

lib/run_printer.py
```diff
@@ -161,13 +165,17 @@ async def dump_run(args):
 
     sleeper = BackoffSleeper(jitter=random.random())
     check_run = get_run_checker()
+    out_file = args.out_file.resolve() if args.out_file else None
+    if out_file and out_file.is_dir():
+        logging.error("--out-file flag expects a file path and not directory")
+        sys.exit(0)
```
angelonakos: Should be `sys.exit(1)`

doc/src/man/litani-dump-run.scdoc
```diff
@@ -62,6 +63,8 @@ _"complete"_ field set to *true*.
 	every second up to the limit given in this flag. If _N_ is *0*, this program
 	will retry indefinitely---this is not recommended.
 
+*--out-file* _F_
+	Output file to print run.json in.
```
karkhaz: I would say something like `Print the run to _F_ instead of stdout`"... I might actually open a separate PR to fix the documentation above, it could be cleaner.

Could you also fix the documentation at the top of the page,

> This program prints the _run.json_ file for an in-progress Litani run to stdout or to the specified output file.

lib/run_printer.py
```diff
@@ -161,13 +165,17 @@ async def dump_run(args):
 
     sleeper = BackoffSleeper(jitter=random.random())
     check_run = get_run_checker()
+    out_file = args.out_file.resolve() if args.out_file else None
```
karkhaz: Please do this error checking in the argparser 🐳

litani
```diff
@@ -260,6 +260,11 @@ def get_args():
             "help":
                 "how many times to retry loading the run in 1 second intervals"
                 " (Default: %(default)s)"
+        }, {
+            "flags": ["--out-file"],
+            "metavar": "F",
+            "type": pathlib.Path,
```
karkhaz: Instead of `pathlib.Path`, I suggest you create a new helper function in this file called `_non_directory_path` which takes a single argument and does the validation you're doing here 🐳. Instead of `logging.error` and `exit`, it should just throw `ValueError` with the error message. If you try it yourself, you should see that it formats a usage/help message for you if you do it this way.

litani
```diff
@@ -260,6 +260,11 @@ def get_args():
             "help":
                 "how many times to retry loading the run in 1 second intervals"
                 " (Default: %(default)s)"
+        }, {
+            "flags": ["--out-file"],
```
karkhaz: This is a user-facing command, can I suggest we also have `"-o"` as a short flag? Please also update the documentation.

karkhaz (changes requested):

litani
karkhaz: (see the example [here](https://docs.python.org/3/library/argparse.html#type))

lib/run_printer.py
ronakfof: Removed this check from this file as suggested by Kareem.

lib/run_printer.py
ronakfof: Resolved

doc/src/man/litani-dump-run.scdoc
ronakfof: Resolved

litani
ronakfof: Resolved

litani
ronakfof: Resolved

lib/run_printer.py
ronakfof: Resolved

litani
```diff
@@ -416,6 +421,12 @@ def non_negative_int(arg):
     return ret
 
 
+def _non_directory_path(arg):
+    if pathlib.Path(arg).exists() and pathlib.Path(arg).resolve().is_dir():
```
karkhaz: just create the `pathlib.Path` object once and assign it to a variable?

litani
```diff
@@ -416,6 +421,12 @@ def non_negative_int(arg):
...
+        raise ValueError("--out-file flag expects a file and not a directory")
```
karkhaz: please include `arg` in the error message

litani
```diff
@@ -416,6 +421,12 @@ def non_negative_int(arg):
...
+    return pathlib.Path(arg).resolve()
```
karkhaz: I don't think we should resolve. If the user gave us a symlink, we should write to the symlink handle.

doc/src/man/litani-dump-run.scdoc
```diff
@@ -23,11 +23,12 @@ litani dump-run - Print the current run as a JSON document to stdout
 
 *litani dump-run*
 	\[*--retries* _N_]
+	\[*--out_file* _F_]
```
karkhaz: `*-o*/*--out-file* _F_`, like in the man page for litani-run-build

karkhaz: Note that it's `out-file` not `out_file`

doc/src/man/litani-dump-run.scdoc
```diff
@@ -62,6 +63,8 @@ _"complete"_ field set to *true*.
 	every second up to the limit given in this flag. If _N_ is *0*, this program
 	will retry indefinitely---this is not recommended.
 
+*--out-file* _F_
```
karkhaz: `*-o* _F_, *--out-file* _F_`

karkhaz (changes requested):

litani
ronakfof: This is generated when we pass a dir - 
```
litani % ./litani dump-run --out-file /tmp/

usage: litani dump-run [-h] [-r N] [-o F]
litani dump-run: error: argument -o/--out-file: invalid _non_directory_path value: '/tmp/'
```

karkhaz (approved):

doc/src/man/litani-dump-run.scdoc
```diff
@@ -22,12 +22,13 @@ litani dump-run - Print the current run as a JSON document to stdout
...
 *litani dump-run*
-	\[*--retries* _N_]
+	\[*-r*/*--retries* _N_]
+	\[*-o*/*--out-file* _F_]
 
 
 # DESCRIPTION
 
-This program prints the _run.json_ file for an in-progress Litani run to stdout.
+This program prints the run.json file for an in-progress Litani run to stdout or to the specified output file.
```
angelonakos: `in-progress` can be omitted. `dump-run` can print the run.json after a Litani build has concluded

lib/run_printer.py
```diff
@@ -112,8 +112,12 @@ def get_run_checker():
 
 
 
-def _exit_success(run):
-    print(json.dumps(run, indent=2))
+def _exit_success(run, out_file):
+    if out_file is not None:
```
angelonakos: What's the difference between `if out_file is not None:` and `if out_file:` ?

ronakfof: I saw it being used [HERE](https://github.com/awslabs/aws-build-accumulator/blob/release/litani#L383-L385) and to be consistent I used the same.

doc/src/man/litani-dump-run.scdoc
ronakfof: Resolved

lib/run_printer.py
ronakfof: Updated

angelonakos: I see. Should it change there as well then?

angelonakos (approved):

PR #114, opened 2022-03-01
Create litani dump-jobs

### LINKED ISSUE #107: Create `litani dump-jobs`
Analagous to `litani dump-run`, it should print a list of jobs to stdout.

Acceptance:
- Implement the command
- It should also have a `-o`/`--out-file` flag to write the list of jobs to a file instead of stdout, some people prefer that to a redirection
- Write a man page, it will probably be similar to dump-run

ronakfof:
Fixes #107

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/run_printer.py
+++ b/lib/run_printer.py
@@ -23,6 +23,7 @@ import os
 import pathlib
 import random
 import signal
+import subprocess
 import sys
 import time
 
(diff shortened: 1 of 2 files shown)


karkhaz:
I think this PR should reuse/refactor the existing code instead of copying it, there should be a way to make the existing code decide whether to dump a whole run or just a list of jobs.

ronakfof:
> I think this PR should reuse/refactor the existing code instead of copying it, there should be a way to make the existing code decide whether to dump a whole run or just a list of jobs.

Instead of having another command, should this be just a flag for `litani dump-run` like `--jobs`?

angelonakos:
> > I think this PR should reuse/refactor the existing code instead of copying it, there should be a way to make the existing code decide whether to dump a whole run or just a list of jobs.
> 
> 
> 
> Instead of having another command, should this be just a flag for `litani dump-run` like `--jobs`?
> 
> 

If adding a new command isn't ideal, then the existing command could be renamed to something like `litani dump`. This new command could take a required `--target` flag, whose value can be either `run` or `jobs` (in fact, the default value could be `run`).

Issue #109, opened 2022-02-28
Cache path to cache directory


karkhaz:
Don't compute it each time.

PR #119, opened 2022-03-03
Cache path to cache directory

ronakfof:
Run get_cache_dir() once and save path to env from where it can be
reused.

*Issue #, if available:*
#109 

*Description of changes:*
Run get_cache_dir() once and save path to `CacheDir.path` variable

### DIFF
--- a/lib/litani.py
+++ b/lib/litani.py
@@ -14,6 +14,7 @@
 
 import asyncio
 import contextlib
+import dataclasses
 import json
 import logging
 import os
@@ -66,6 +67,48 @@ class AcquisitionFailed(Exception):
 class TimeoutExpired(Exception):
     pass
 
+
+@dataclasses.dataclass
+class CacheDir:
+    cache_dir_path: pathlib.Path = None
+
+    @staticmethod
+    def get(lookup_path=os.getcwd()):
+        if CacheDir.cache_dir_path:
+            return CacheDir.cache_dir_path
+        def cache_pointer_dirs():
+            current = pathlib.Path(lookup_path).resolve(strict=True)
+            yield current
+            while current.parent != current:
+                current = current.parent
+                yield current
+
+        for possible_dir in cache_pointer_dirs():
+            logging.debug(
+                "Searching for cache pointer in %s", possible_dir)
+            possible_pointer = possible_dir / CACHE_POINTER
+            try:
+                if possible_pointer.exists():
+                    logging.debug(
+                        "Found a cache pointer at %s", possible_pointer)
+                    with open(possible_pointer) as handle:
+                        pointer = handle.read().strip()
+                    possible_cache = pathlib.Path(pointer)
+                    if possible_cache.exists():
+                        logging.debug("cache is at %s", possible_cache)
+                        CacheDir.cache_dir_path = possible_cache
+                        return possible_cache
+                    logging.warning(
+                        "Found a cache file at %s pointing to %s, but that "
+                        "directory does not exist. Continuing search...",
+                        possible_pointer, possible_cache)
+            except PermissionError:
+                pass
+
+        logging.error(
+            "Could not find a pointer to a litani cache. Did you forget "
+            "to run `litani init`?")
+        raise FileNotFoundError
 
 
 class LockableDirectory:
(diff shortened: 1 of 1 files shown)


lib/litani.py
```diff
@@ -168,6 +168,7 @@ def cache_pointer_dirs():
                 possible_cache = pathlib.Path(pointer)
                 if possible_cache.exists():
                     logging.debug("cache is at %s", possible_cache)
+                    os.environ["CACHE_DIR"] = str(possible_cache.resolve())
```
karkhaz: I don't think we should store this in an environment variable, there's no need for subprocesses to know about it.

Suggest that you create a new class called CacheDir with a static method called `get`. This method checks whether `CacheDir.path` (a static variable) is set, if so it returns it, if not it does the computation here.

There's also no need to resolve the path before returning it.

karkhaz (changes requested):

lib/litani.py
```diff
@@ -168,6 +176,7 @@ def cache_pointer_dirs():
                 possible_cache = pathlib.Path(pointer)
                 if possible_cache.exists():
                     logging.debug("cache is at %s", possible_cache)
+                    CacheDir.path = possible_cache
```
angelonakos: Isn't the following more typical of an instantiation?

```python
cache_dir = CacheDir(path=possible_cache)
```

In fact, can't you just say `return CacheDir(path=possible_cache) and remove both this line and the line beneath it?

lib/litani.py
```diff
@@ -184,7 +193,7 @@ def cache_pointer_dirs():
 
 def get_cache_dir(path=os.getcwd()):
     try:
-        return _get_cache_dir(path)
+        return CacheDir.get() or _get_cache_dir(path)
```
angelonakos: Do you think the following is more clear? 

`return CacheDir(path=path) or _get_cache_dir(path)`

angelonakos (changes requested):

lib/litani.py
ronakfof: We have to return the `path` and not the object.

lib/litani.py
ronakfof: We have to return `path` and not object.

lib/litani.py
karkhaz: I like this implementation! Can I suggest one last change: move the current implementation of `_get_cache_dir` into `get`. That is, `get` should do `if self.path: return path` and then the rest of the method should be `_get_cache_dir` trying to set `path` and then returning it.

Then this line can be just `return CacheDir.get()`

karkhaz (changes requested):

lib/litani.py
angelonakos: Right. Doesn't the following return the path attribute?

`return CacheDir(path=path).get() or _get_cache_dir(path)`

lib/litani.py
angelonakos: Right. Doesn't the following return the path attribute?

`return CacheDir(path=possible_cache).path`

ronakfof: Yes. Made some changes along with this.

lib/litani.py
ronakfof: Yes. Made some changes along with this.

ronakfof: Resolved

lib/litani.py
```diff
@@ -147,44 +190,10 @@ async def try_acquire_wait(self, timeout=0):
...
-    logging.error(
-        "Could not find a pointer to a litani cache. Did you forget "
-        "to run `litani init`?")
-    raise FileNotFoundError
-
-
 def get_cache_dir(path=os.getcwd()):
     try:
-        return _get_cache_dir(path)
+        cache_dir = CacheDir(lookup_path=path)
```
nrdg42: Just as a matter of personal preference I think the construction of this object ought to go outside of the try/except because you shouldn't expect that constructing an object will raise an exception.

nrdg42: Also, is there a reason that we pass the lookup_path into the constructor rather than just to the `get()` method?

ronakfof: > Just as a matter of personal preference I think the construction of this object ought to go outside of the try/except because you shouldn't expect that constructing an object will raise an exception.

Makes sense. Will update.


> Also, is there a reason that we pass the lookup_path into the constructor rather than just to the `get()` method?

Because previously we were passing a `path` var to the `_get_cache_dir` function which was set to `os.getcwd()` as default.

nrdg42: I don't understand. Can we not now pass a `path` var into `get()` that is set to `os.getcwd()` as default? Why does it need to go in the constructor?

ronakfof: Resolved

lib/litani.py
```diff
@@ -67,6 +68,47 @@ class TimeoutExpired(Exception):
     pass
 
 
+@dataclasses.dataclass
+class CacheDir:
+    cache_dir_path: pathlib.Path = None
+
+    def get(self, lookup_path=os.getcwd()):
```
karkhaz: Both this method and the `cache_dir_path` field should be static. The whole point is that we shouldn't need to create a `CacheDir` object every time we want to retrieve the path, since that will entail re-traversing the directory tree. We only want this method called once during the entire run of the program.

lib/litani.py
```diff
@@ -147,44 +189,10 @@ async def try_acquire_wait(self, timeout=0):
...
-        "Could not find a pointer to a litani cache. Did you forget "
-        "to run `litani init`?")
-    raise FileNotFoundError
-
-
 def get_cache_dir(path=os.getcwd()):
+    cache_dir = CacheDir()
     try:
-        return _get_cache_dir(path)
+        return cache_dir.get(lookup_path=path)
```
karkhaz: `return CacheDir.get(path)`

karkhaz (changes requested):

ronakfof:
I printed how many times the `cache_dir` is being calculated and it seems like it's 1 per `add-job` instead of once per whole run? Is that expected?

karkhaz (approved):

angelonakos (approved):

PR #126, opened 2022-03-07
Rebuild run graph after transform jobs

### LINKED ISSUE #105: dicts emitted from litani-transform-jobs should only contain command-line switches
Currently, litani add-job emits a list of dicts. The schema of these dicts is exactly the `wrapper_arguments` schema in run.json. However, this contains a few fields that the user shouldn't have to deal with, like job UUIDs and status files.

We should remove these fields from the output entirely.

In fact, the current implementation distinguishes between jobs that get added, transformed, and deleted. There's no need to do this at all. Just wipe the entire status directory and re-write a bunch of new jobs, with new UUIDs even.

nrdg42:
Transform jobs will only output dicts with keys that are command-line
switches. After receiving new user jobs, Litani will discard the
previous run graph and build a new one

This fixes #105

### DIFF
--- a/lib/add_job.py
+++ b/lib/add_job.py
@@ -1,0 +1,55 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License
+
+import json
+import logging
+import sys
+import uuid
+
+from lib import litani
+
+
+async def add_job(job_dict):
+    cache_file = litani.get_cache_dir() / litani.CACHE_FILE
+    with open(cache_file) as handle:
+        cache_contents = json.load(handle)
+    if job_dict["ci_stage"] not in cache_contents["stages"]:
+        valid_stages = "', '".join(cache_contents["stages"])
+        logging.error(
+            "Invalid stage name '%s' was provided, possible "
+            "stage names are: '%s'", job_dict["ci_stage"], valid_stages)
+        sys.exit(1)
+
+    jobs_dir = litani.get_cache_dir() / litani.JOBS_DIR
+    jobs_dir.mkdir(exist_ok=True, parents=True)
+
+    if job_dict["phony_outputs"]:
+        if not job_dict["outputs"]:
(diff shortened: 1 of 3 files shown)


lib/add_job.py
```diff
@@ -0,0 +1,42 @@
+import json
```
karkhaz: License header please

lib/add_job.py
```diff
@@ -0,0 +1,42 @@
...
+import logging
+import sys
+import uuid
+
+from lib import litani
+
+
+async def add_job(job_dict):
```
karkhaz: This seems like it's been moved without modifications, could it be a different commit?

lib/transform_jobs.py
```diff
@@ -28,76 +29,36 @@ class _JobsTransformer:
...
+
+    async def _add_new_jobs(self, user_jobs):
+        for job in user_jobs:
+            await lib.add_job.add_job(job)
+
+
+    def _delete_old_jobs(self):
         for uuid in self.old_uuids:
             job_file = self.jobs_dir / ("%s.json" % uuid)
+            logging.info("Discarding job %s", job_file.name)
```
karkhaz: Since we're always discarding all old jobs, I no longer think this message needs to be here, it's not that helpful.

lib/transform_jobs.py
```diff
@@ -28,76 +29,36 @@ class _JobsTransformer:
...
             try:
-                with open(job_file) as handle:
-                    old_job = json.load(handle)
-                    uuid = old_job["job_id"]
+                os.unlink(job_file)
             except FileNotFoundError:
-                logging.warning(
-                    "file for job %s disappeared; not updating", uuid)
+                logging.error("Unable to delete %s", job_file.name)
```
karkhaz: Again, I think this is no longer notable, no need to log.

lib/transform_jobs.py
```diff
@@ -28,76 +29,36 @@ class _JobsTransformer:
...
-
 
 
 def _print_jobs(job_paths):
     out = []
     for job in job_paths:
         with open(job) as handle:
-            out.append(json.load(handle))
+            job_dict = json.load(handle)
+            for key in ("job_id", "status_file", "subcommand"):
```
karkhaz: You're not adding `"subcommand"` back in when you restore the jobs, it looks like

karkhaz (changes requested):
Looking forward to this!

lib/transform_jobs.py
nrdg42: Do we use the `subcommand` key? Should it be the same as it was before transform jobs or should it be set to `transform-jobs`?

angelonakos: > Do we use the subcommand key? 

Not sure. I haven't checked.

> Should it be the same as it was before transform jobs or should it be set to transform-jobs?

I would prefer if it is set to `transform-jobs` to reflect the fact that the subcommand has changed.

karkhaz (approved):

angelonakos (approved):

PR #130, opened 2022-03-14
Add subcommands for get-jobs and set-jobs

nrdg42:
*Issue #, if available:*

*Description of changes:*
Add `litani get-jobs` and `litani set-jobs` subcommands and refactor `litani transform-jobs` to use those functions. This allows litani to be more flexible around job creation and helps clean up `litani transform-jobs`. 


litani
```diff
@@ -281,9 +280,23 @@ def get_args():
         flags = arg.pop("flags")
         graph_pars.add_argument(*flags, **arg)
 
+    get_jobs_pars = subs.add_parser("get-jobs", help="print jobs")
```
karkhaz, line 175: One request (specifically from @markrtuttle). Can both of these commands take a file? `--from-file` for `set-jobs` and `--out-file` for `get-jobs`. (I suggest the short flag is `-f` in both cases).

A very clean implementation could be:
* In the argument parser, set `default` to `sys.stdin` or `sys.stdout`
* `type` should be a function that takes a path to a file, opens it for reading/writing, and returns the handle. (So use `open` as a straight function call, without using `with`). For `set-jobs`, if the file to be read doesn't exist then raise.
* This means that the value of `args.from_file` will always be a handle (whether a real file or stdin) and you can just uniformly call `read` on it to read the JSON, or `print(..., file=args.out_file)`.

litani
```diff
@@ -281,9 +280,23 @@ def get_args():
...
+    get_jobs_pars.set_defaults(func=lib.jobs.print_jobs)
+
+    set_jobs_pars = subs.add_parser("set-jobs",
+        help="set jobs to run")
+    for arg in [{
+            "flags": ["-j", "--jobs"],
+            "help": "set jobs to run by passing in a json list of job dicts",
```
karkhaz: I think "set jobs to run" is a bit redundant since that's what the command does. Can I suggest something like
```
read a list of jobs from a JSON file (default: read from stdin)
```
```
write the list of jobs to a JSON file (default: write to stdout)
```

litani
```diff
@@ -642,7 +655,12 @@ async def init(args):
 
 
 async def add_job(args):
-    await lib.add_job.add_job(vars(args))
+    await lib.jobs.add_job(vars(args))
+
+
+async def set_jobs(args):
+    job_list = json.loads(args.jobs)
```
karkhaz: I'd like to get a bit better about error handling in Litani since it's pretty confusing when it crashes, most users have no idea what to do.

Could you catch `JSONDecodeError` here, print an error message and exit (no stack trace)?

You probably also want to catch `FileNotFoundError` when you open the file in the first place (as I suggested, in the `type` argument of the argparser itself). No need here, because at this point you already have an open file handle.

lib/jobs.py
```diff
@@ -53,3 +54,49 @@ async def add_job(job_dict):
...
+                job_dict.pop(key)
+            out.append(job_dict)
+    return out
+
+
+def print_jobs(_=None):
+    jobs = get_jobs()
+    print(json.dumps(jobs, indent=2))
+    sys.stdout.flush()
+    os.close(sys.stdout.fileno())
```
karkhaz, line 240: I realise I didn't write a comment here originally. Would you mind explaining that `sys.stdout.close()` doesn't work because it only closes the Python object representing the file, but the underlying file handle (in the C stdlib) for std{in,out,err} would remain open, which is not what we want. We absolutely must close the handle properly because of `transform-jobs`; the clients of that command would be waiting for End-of-File before they start parsing.

karkhaz, line 240: I realise I didn't write a comment here originally. Would you mind explaining that `sys.stdout.close()` doesn't work because it only closes the Python object representing the file, but the underlying file handle (in the C stdlib) for std{in,out,err} would remain open, which is not what we want.

lib/jobs.py
```diff
@@ -53,3 +54,49 @@ async def add_job(job_dict):
...
         print(json.dumps(job_dict, indent=2), file=handle)
+
+
+def get_jobs():
+    out = []
+    jobs_dir = litani.get_cache_dir() / litani.JOBS_DIR
+    for job in jobs_dir.iterdir():
+        with open(job) as handle:
+            job_dict = json.load(handle)
+            for key in ("job_id", "status_file", "subcommand"):
```
karkhaz: Suggest:
* This tuple is a global variable in this module (but private)
* You reference it here through the variable name
* In `add-job`, just before you write the file, raise `AssertionError` if any of those keys in the tuple are not part of the dict that you're about to write, as that suggests a programming error.

This guards against future mistakes when changing this code.

doc/src/man/litani-get-jobs.scdoc
```diff
@@ -0,0 +1,45 @@
...
+
+# DESCRIPTION
+
+This program prints the list of jobs that have been added so far in JSON
+format. Each job in the list will follow the same schema as the elements of the
+*["jobs"]["wrapper_arguments"]* key in the litani _run.json_ file; see
+*litani-run.json(5)* for the schema.
+
+
+# MOTIVATION & BEST PRACTICES
```
karkhaz, line 36: A few things worth mentioning IMHO:
* This command may be run at any point after running `litani init`---either before calling `litani run-build`, during the run, or after run-build has terminated.
* Multiprocess safety should be a whole other section. This command is not safe to run concurrently with `set-jobs`, because `set-jobs` might delete a JSON file while this command reads it. This command is safe to run concurrently with `add-jobs` and `run-build` because new job files get written atomically (and `run-build`/`exec` write files atomically, but don't mention `exec`), so this command will either read the whole file or miss it completely.
* Compared to `dump-run`, this command will always return immediately if it's successful; `dump-run` has to lock the build while it runs. So if you just need the list of jobs and not other metadata of the run, this command might be better for you. On the other hand: the run contains the Litani version number, which you might want to use for checking that a certain key is part of the schema, so in general it can help you write more robust client code.
* Compared to `transform-jobs`, this command is obviously useful if you just want to read the jobs and not write them... but a usage difference is that clients of `transform-jobs` have to access the stdin and stdout of a subprocess, while this command can use input/output files and can be used in a Unix pipeline, which may be more natural.
* Can I suggest a short (for now) EXAMPLES section, maybe even just one example showing a pipeline? I haven't tested this but
```
# Pipeline
litani get-jobs | jq '[.[] | (.wrapper_arguments.command = "echo " + .wrapper_arguments.command)]' | litani set-jobs

# Sequence
litani get-jobs --out-file /tmp/original.json
jq '[.[] | (.wrapper_arguments.command = "echo " + .wrapper_arguments.command)]' < /tmp/original.json > /tmp/transformed.json
litani set-jobs --from-file /tmp/transformed.json
```
will prepend `echo ` to each command.

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,44 @@
...
+
+# NAME
+
+litani set-jobs - Set the list of jobs to be run
+
+
+# DESCRIPTION
+
+This program reads a list of jobs as a JSON list. Each job in the list will
+be a dict with keys being the flags to *litani add-job(1)*. This list of jobs
```
karkhaz: > keys being the flags to *litani add-job(1)*

Not exactly, dashes must be replaced by underscore unfortunately.

karkhaz (changes requested):
This is an excellent implementation, thank you very much!

The only additional thing this needs is some tests. Can I suggest that you use a similar mechanism in the test suite that we use for `transform_jobs`? That is: if one of the files under `test/e2e` contains a method called `get_jobs`, then `litani get-jobs` gets invoked and the result is returned; and if there's a method called `set_jobs` that takes a single argument, then again the command gets run.

See for example `test/e2e/tests/transform_delete_job.py`, the `OPERATIONS` dict in `test/e2e/run`, and where it says `if add_transform_jobs` in `test/run`.

Suggestions for tests: get-jobs and check that they're what you expect; do set-jobs followed by get-jobs to see that the result is what you expect; run get-jobs in the edge case where the list of jobs is empty, and similarly for set-jobs; etc.

litani
karkhaz: (When I said 'raise', I specifically mean raise `argparse.ArgumentError`).

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+; permissions and limitations under the License.
+
+
+# NAME
+
+litani set-jobs - Set the list of jobs to be run
+
+# SYNOPSIS
+
+*litan set-jobs*
```
karkhaz: `litani`

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+	\[*-j*/*--jobs* _J_]
```
karkhaz: This is a nice idea! Can I suggest `--from-string` instead, that seems more descriptive to me? I wouldn't mind getting a second opinion on that.

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+
+# EXAMPLES
+
+Below is an example which demonstrates how to combine *litani get-jobs* and
+*litani set-jobs* to prepend echo to each command that has been added to a
+litani run either using a pipeline or a sequence
+
+```
+# Pipeline
+litani get-jobs | jq '[.[] | (.wrapper_arguments.command = "echo " + .wrapper_arguments.command)]' | litani set-jobs
```
karkhaz: BTW does this actually work? I didn't test it last week. I think `get-jobs` prints out the value of the `"wrapper_arguments"` key so actually the JQ expression shouldn't include `.wrapper_arguments`?

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+be more natural for use in a Unix pipeline.
+
+
+# EXAMPLES
+
+Below is an example which demonstrates how to combine *litani get-jobs* and
+*litani set-jobs* to prepend echo to each command that has been added to a
+litani run either using a pipeline or a sequence
+
+```
```
karkhaz, line 55: This is scdoc not markdown... you can read the syntax with `man 5 scdoc`. Code blocks are indented, not surrounded by backticks

lib/jobs.py
```diff
@@ -51,5 +54,73 @@ async def add_job(job_dict):
...
+    # file handle in the C stdlib would remain open, so this is how we need to
+    # close it for transform_jobs
+    sys.stdout.flush()
+    os.close(sys.stdout.fileno())
+
+
+def _delete_jobs():
+    jobs_dir = litani.get_cache_dir() / litani.JOBS_DIR
+
+    if not jobs_dir.exists():
```
karkhaz: This is just a nit, but I think the more Pythonic thing to do is:
1. Don't check that the directory exists
2. Wrap the entire for loop in a try-catch for `FileNotFoundError`. That catches all the following cases:
  * The directory didn't exist initially
  * Some file within the directory gets deleted during the iteration
  * The entire directory gets deleted during the iteration

I think the non-pythonic implementation doesn't protect against the third item, because you want `jobs_dir.iterdir()` to be inside the try-catch in case *that* throws.

lib/jobs.py
```diff
@@ -51,5 +54,73 @@ async def add_job(job_dict):
...
+async def set_jobs(job_list):
+    _delete_jobs()
+    for job in job_list:
+        job["subcommand"] = "add-job"
+        await add_job(job)
+
+
+def _read_jobs():
+    in_text = sys.stdin.read()
+    return json.loads(in_text)
```
karkhaz: Please catch JSONDecodeError here

litani
```diff
@@ -281,9 +280,39 @@ def get_args():
         flags = arg.pop("flags")
         graph_pars.add_argument(*flags, **arg)
 
+    get_jobs_pars = subs.add_parser("get-jobs", help="print jobs")
+    for arg in [{
+        "flags": ["-f", "--out-file"],
+        "help": "write the list of jobs to a JSON file, default: stdout",
+        "default": sys.stdout,
+        "metavar": "F",
+        "type": argparse.FileType("w")
```
karkhaz, line 181: Ooh nice!

litani
```diff
@@ -281,9 +280,39 @@ def get_args():
...
+    for arg in [{
+            "flags": ["-j", "--jobs"],
+            "help": "read a list of jobs from a JSON string",
+            "metavar": "J"
+    }, {
+            "flags": ["-f", "--from-file"],
+            "help": "read a list of jobs from a JSON file, defaut: stdout",
+            "metavar": "F",
+            "default": sys.stdin,
+            "type": open
```
karkhaz: This is also nice, but... If the file doesn't exist, this will throw `FileNotFoundError` and print a stack trace, right? Whereas if you implemented a custom function that raised `ValueError` or `argparse.ArgumentError`, the argparser will actually print a nice help text instead...

test/e2e/run
```diff
@@ -98,6 +104,28 @@ def transform_jobs(litani, run_dir, mod):
         sys.exit(1)
 
 
+def get_jobs(litani, run_dir, mod):
+    args = mod.get_check_get_jobs_args()
+
+    os.chdir(run_dir)
```
karkhaz: As far as I can tell it's currently safe to change the working directory of this script without changing back... but I still think it's unnecessary and will lead to difficult-to-track-down bugs in the future. Can I recommend that you use the `cwd` argument to `subprocess.{run,check_output}` instead?

test/e2e/run
```diff
@@ -98,6 +104,28 @@ def transform_jobs(litani, run_dir, mod):
...
+
+    mod.check_get_jobs(jobs)
+
+    for job in mod.get_post_check_jobs():
+        run_litani(
+            litani, "add-job", *job.get("args", []), **job.get("kwargs", {}))
+
+
+def set_jobs(litani, run_dir, mod):
+    os.chdir(run_dir)
```
karkhaz, line 123: Maybe `run_litani` should have a `cwd` keyword argument before `*args` and `**kwargs`... so that you can pass in what directory Litani runs in.

test/e2e/tests/get_jobs_multiple_jobs.py
```diff
@@ -0,0 +1,87 @@
...
+
+def get_check_get_jobs_args():
+    return {}
+
+
+
+def check_get_jobs(jobs_json):
+    jobs = json.loads(jobs_json)
+    print(json.dumps(jobs, indent=2))
+    if len(jobs) != 3:
```
karkhaz, line 59: Can the rest of this method be
```
if set([j["command"] for j in jobs]) != {"echo foo", "echo bar", "echo baz"}:
   return False
```

test/e2e/tests/get_jobs_multiple_jobs.py
```diff
@@ -0,0 +1,87 @@
...
+        "echo baz",
+        "echo foo",
+    ]
+
+    for job in jobs:
+        expected_commands.remove(job["command"])
+
+    if expected_commands != []:
+        logging.error("Expected jobs to contain %s commands", expected_commands)
+        sys.exit(1)
```
karkhaz: The other `check_*` test commands conventionally return `True` if the test is successful and `False` otherwise. Is there a good reason to abort the program like this? I think the error message should also print the actual result not just the expected one.

test/e2e/tests/get_jobs_multiple_jobs.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def get_check_get_jobs_args():
+    return {}
+
+
+
+def check_get_jobs(jobs_json):
+    jobs = json.loads(jobs_json)
+    print(json.dumps(jobs, indent=2))
```
karkhaz: Why print from a test? If it needs to be logged can we use `logging`?

test/e2e/tests/get_jobs_no_jobs.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_check_get_jobs_args():
+    return {}
+
+
+def check_get_jobs(jobs_json):
+    jobs = json.loads(jobs_json)
+    if jobs != []:
+        logging.error("Not expecting to get any jobs, got %s", jobs)
+        sys.exit(1)
```
karkhaz: return False

test/e2e/tests/get_jobs_no_jobs.py
```diff
@@ -0,0 +1,61 @@
...
+            "pipeline": "foo",
+        }
+    }]
+
+
+def get_run_build_args():
+    return {}
+
+
+def check_run(run):
```
karkhaz, line 61: To be honest maybe this should just be `return True`? It's nice to just test a single thing in a single test. You want to emphasize the fact that the important check here is actually the `get_jobs` one.

test/e2e/tests/get_jobs_single_job.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def check_get_jobs(jobs_json):
+    jobs = json.loads(jobs_json)
+    print(json.dumps(jobs, indent=2))
+    if len(jobs) != 1:
+        logging.error("Expected to get 1 job, got %s", len(jobs))
+        sys.exit(1)
+
+    expected_job = {
```
karkhaz: Make this global and then `get_jobs()` can be `return [{"kwargs": EXPECTED_JOB}]`?

test/e2e/tests/get_jobs_single_job.py
```diff
@@ -0,0 +1,85 @@
...
+    }
+    job = jobs[0]
+    for key in expected_job.keys():
+        if job.get(key) != expected_job[key]:
+            logging.error(
+                "Expected job to have key-value pair '%s': '%s'",
+                key, expected_job[key])
+            sys.exit(1)
+
+    removed_keys = [
```
karkhaz, line 58: Would it be a good idea to just re-use `lib.jobs._PRIVATE_JOB_FIELDS` here?

test/e2e/tests/set_multiple_jobs.py
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+
+def get_jobs():
+    return []
+
+
+def get_set_jobs_args():
+    jobs = [{
```
karkhaz, line 36: OK, this highlights something I hadn't noticed before.

It seems like it's necessary for the user to supply *all* of these keys, right? They can't for example omit `"outputs"` even if the value is `None`?

That seems (1) unergonomic and (2) we would break everybody's scripts every time we add a new key to the schema, because if somebody's script is doing `set-jobs` with a certain set of keys that doesn't include the new key that we just introduced, then something somewhere will crash.

I think every flag that has a default or optional value in the command line needs to also be that way in the JSON. Namely the only required keys in the set-jobs dict should be command, pipeline, ci_stage.

I realise that this will make the implementation a lot more complicated :'( But I don't think we can merge this in like this.

test/run
```diff
@@ -92,12 +107,10 @@ def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
         if test_file.name in ["__init__.py", "__pycache__"]:
             continue
 
-        add_transform_jobs = False
-        with open(test_file) as handle:
-            for line in handle:
-                if line.strip().startswith("def transform_jobs("):
-                    add_transform_jobs = True
-                    break
+        add_transform_jobs = has_line(test_file, "def transform_jobs(")
```
karkhaz, line 111: Nice

karkhaz (changes requested):

test/e2e/tests/get_jobs_multiple_jobs.py
karkhaz: Well, even `return set([j["command"] for j in jobs]) == {"echo foo", "echo bar", "echo baz"}` is shorter, although I do think it's good to print an error message with the actual values.

litani
```diff
@@ -281,9 +280,39 @@ def get_args():
...
+    set_jobs_pars = subs.add_parser("set-jobs",
+        help="set jobs to run")
+    set_jobs_mutually_exclusive = set_jobs_pars.add_mutually_exclusive_group()
+    for arg in [{
+            "flags": ["-j", "--jobs"],
+            "help": "read a list of jobs from a JSON string",
+            "metavar": "J"
+    }, {
+            "flags": ["-f", "--from-file"],
+            "help": "read a list of jobs from a JSON file, defaut: stdout",
```
karkhaz: `default` is spelled incorrectly, and also this is `stdin` not `stdout`

doc/src/man/litani-set-jobs.scdoc
nrdg42: The man pages tell me to block off literal text with backticks which automatically indent code blocks by one level

lib/jobs.py
nrdg42: Why would the error message that I would display and then exit with be more helpful for the user than the json Decode error and a stack trace?

test/e2e/run
nrdg42: Why don't we use that anywhere else in the file?

test/e2e/run
nrdg42: you have a cwd test that passes in a cwd keyword already

test/e2e/tests/get_jobs_no_jobs.py
nrdg42: Then why write an end to end test instead of a unit test?

test/e2e/tests/get_jobs_single_job.py
nrdg42: Do I have access to the lib module?

test/e2e/tests/get_jobs_multiple_jobs.py
nrdg42: It's not clear to me how we determine if a test was successful or not

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def _delete_jobs():
+    jobs_dir = litani.get_cache_dir() / litani.JOBS_DIR
+
+    try:
+        for job_file in jobs_dir.iterdir():
+            os.unlink(job_file)
+    except FileNotFoundError:
+        pass
```
angelonakos, line 250: Should a logging statement be added?

litani
```diff
@@ -281,9 +172,39 @@ def get_args():
...
+        "help": "write the list of jobs to a JSON file, default: stdout",
+        "default": sys.stdout,
+        "metavar": "F",
+        "type": argparse.FileType("w")
+    }]:
+        flags = arg.pop("flags")
+        get_jobs_pars.add_argument(*flags, **arg)
+    get_jobs_pars.set_defaults(func=lib.jobs.print_jobs)
+
+    set_jobs_pars = subs.add_parser("set-jobs",
```
angelonakos, line 187: Is indentation off?

nrdg42: what makes you say that?

lib/jobs.py
karkhaz: No need. This is pretty expected if more than one of these are running concurrently

litani
angelonakos: Disregard earlier comment. Indentation's fine.

test/run
```diff
@@ -153,8 +175,49 @@ def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
...
+                    --operation set-jobs"""),
+                pipeline=f"End-to-end: {test_file.stem}",
+                ci_stage="test",
+                description=f"{test_file.stem}: set jobs",
+                inputs=add_jobs_output,
+                phony_outputs=set_jobs_output,
+                cwd=run_dir)
+            optional_outputs.append(set_jobs_output)
+
+        run_build_input = add_jobs_output
```
angelonakos: Should this be `run_build_input = str(uuid.uuid4())` like before?

lib/jobs.py
angelonakos:
```python
for job_file in jobs_dir.iterdir():
    try:
        os.unlink(job_file)
    except FileNotFoundError:
        pass
```
?

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+*litani set-jobs*
+	\[*-f*/*--from-file* _F_]
+	\[*-s*/*--from-string* _S_]
+
+
+# DESCRIPTION
+
+This program reads a list of jobs as a JSON list. Each job in the list will be
+a dict with keys being the flags to *litani add-job(1)* with dashes replaced by
+underscores. This list of jobs will replace those in the current build graph.
```
karkhaz: > with dashes replaced by underscores

...is there a good reason for this? I wonder if we just let users use dashes and replace them ourselves?

doc/src/man/litani-set-jobs.scdoc
```diff
@@ -0,0 +1,77 @@
...
+# SYNOPSIS
+
+*litani set-jobs*
+	\[*-f*/*--from-file* _F_]
+	\[*-s*/*--from-string* _S_]
+
+
+# DESCRIPTION
+
+This program reads a list of jobs as a JSON list. Each job in the list will be
```
karkhaz: Suggest a more imperative/active tone here? `Each job in the list should be` and `This program will replace all previously-added jobs with the jobs in the user-provided list.`

doc/src/man/litani-set-jobs.scdoc
karkhaz: Ah, thanks. I I was confused... just saw that the man pages I wrote do use backticks, that's a relief. Thanks.

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+        if value is None:
+            continue
+        if isinstance(value, bool):
+            if value:
+                cmd.append(switch)
+            continue
+
+        cmd.append(switch)
+        if isinstance(value, list):
+            logging.error(value)
```
karkhaz: why?

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+            cmd.extend(str(v) for v in value)
+        else:
+            cmd.append(str(value))
+    return cmd
+
+
+def fill_job(job):
```
karkhaz, line 155: Nice!!

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+        for job_file in jobs_dir.iterdir():
+            os.unlink(job_file)
+    except FileNotFoundError:
+        pass
+
+
+async def set_jobs(job_list):
+    _delete_jobs()
+    for job in job_list:
+        logging.error(job)
```
karkhaz: why?

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+        filled_job = fill_job(job)
+        filled_job["subcommand"] = "add-job"
```
karkhaz, line 257: Why isn't this line in `add_job()`? Filling in the `job_id` and `status_file` fields are already in that method...

litani
```diff
@@ -281,9 +172,39 @@ def get_args():
...
+        help="set jobs to run")
+    set_jobs_mutually_exclusive = set_jobs_pars.add_mutually_exclusive_group()
+    for arg in [{
+            "flags": ["-j", "--jobs"],
```
karkhaz: `-j` here but `-s`/`--from-string` in the man page

lib/jobs.py
```diff
@@ -0,0 +1,276 @@
...
+            for key in _PRIVATE_JOB_FIELDS:
+                job_dict.pop(key)
+            out.append(job_dict)
+    return out
+
+
+async def print_jobs(args=None):
+    jobs = await get_jobs()
+
+    out_file = sys.stdout
```
karkhaz, line 230: I don't think you need this code? The argparser already takes care of it, you've set the default to be stdout and the type to be a file to write to

litani
```diff
@@ -281,9 +172,39 @@ def get_args():
...
+    }, {
+            "flags": ["-f", "--from-file"],
+            "help": "read a list of jobs from a JSON file, default: stdin",
+            "metavar": "F",
+            "default": sys.stdin,
+            "type": argparse.FileType("r")
+    }]:
+        flags = arg.pop("flags")
+        set_jobs_mutually_exclusive.add_argument(*flags, **arg)
+    set_jobs_pars.set_defaults(func=set_jobs)
```
karkhaz, line 203: Why isn't this `lib.jobs.set_jobs`?

litani
```diff
@@ -642,7 +563,20 @@ async def init(args):
 
 
 async def add_job(args):
-    await lib.add_job.add_job(vars(args))
+    await lib.jobs.add_job(vars(args))
+
+
+async def set_jobs(args):
```
karkhaz, line 569: Oh wait, it looks like you have an identical implementation of this in `lib.jobs`?

test/e2e/run
karkhaz: I just had a look. You're right, but actually the code above looks wrong in that case... I think you need to chdir before you load methods from the module (`mod`), right? That's what all the other methods do. So this actually isn't about the subprocess, this is about loading the python module correctly.

So then I think just swap the two above lines, that should be okay?

test/e2e/run
karkhaz: Thanks, disregard

test/e2e/tests/get_jobs_single_job.py
karkhaz: You're right, that won't work.

test/run
karkhaz: I think this should actually be:
1. Instead of `optional_outputs`, call it `run_build_dependencies`
2. Initially its value is `[add_jobs_output]`
3. Append the optional values as you're doing

Then just use `" ".join(run_build_dependencies)` as the input, no need for the if statement. Then run-build takes at least `add_jobs_output` as an input plus possibly some other inputs. Does that work?

test/run
```diff
@@ -153,8 +175,49 @@ def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
...
+                command=collapse(f"""
+                    {e2e_test_dir / 'run'}
+                    --test-file {test_file}
+                    --litani {litani}
+                    --run-dir {run_dir}
+                    --operation set-jobs"""),
+                pipeline=f"End-to-end: {test_file.stem}",
+                ci_stage="test",
+                description=f"{test_file.stem}: set jobs",
+                inputs=add_jobs_output,
```
karkhaz, line 213: Note that `get_jobs` and `set_jobs` are not currently ordered with respect to each other. They both depend on the same file, so they're going to potentially run at the same time.

karkhaz (changes requested):

lib/jobs.py
nrdg42: I had something like that before, but jobs_dir.iterdir() may throw a FileNotFound error if the jobs directory doesn't exist. Kareem suggested this implementation

doc/src/man/litani-set-jobs.scdoc
nrdg42: I think with my new implementation of set-jobs it shouldn't really matter if users pass in dashes or underscores

lib/jobs.py
nrdg42: the logging statement was probably a debugging thing I forgot to take out

lib/jobs.py
nrdg42: Because `subcommand` is something that is automatically added by `argparse`, and I didn't want to modify that

test/run
nrdg42: Is that an issue? I thought that was how it is supposed to work

lib/jobs.py
nrdg42: debugging probably

lib/jobs.py
nrdg42: I call `print_jobs()` from `transform_jobs()` as well, which doesn't pass in any arguments

litani
nrdg42: Because I have a helper function in the litani file which loads the list of jobs based on how the user gave them to us and then calls the `lib.jobs.set_jobs()` function and passes in the list of jobs.

litani
nrdg42: I don't know what you're referring to? The one in lib.jobs expects to take a list of jobs, so this function parses that list of jobs and passes it into that function. Definitely not an identical implementation. There's also `_read_jobs()` which is used by `transform_jobs()` that reads jobs from stdin that I could use, but it doesn't have access to the args

test/e2e/run
nrdg42: I can do that but I haven't had any issues running the tests with the lines in the opposite order so I'm not completely convinced that is whats going on

test/run
nrdg42: That messes with the order because we want to ensure that add_jobs runs before the "optional" outputs right? If we do it the way you're suggesting won't litani try to schedule add_jobs to run at the same time as the other tests? That means that things like transform_jobs won't have the proper jobs added to transform

test/run
```diff
@@ -153,8 +175,49 @@ def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
...
+            get_jobs_output = str(uuid.uuid4())
+            litani_add(
+                litani, counter,
+                command=collapse(f"""
+                    {e2e_test_dir / 'run'}
+                    --test-file {test_file}
+                    --litani {litani}
+                    --run-dir {run_dir}
+                    --operation get-jobs"""),
+                pipeline=f"End-to-end: {test_file.stem}",
```
ronakfof: Please update to `pipeline=f"e2e_{test_file.stem}",`

test/run
```diff
@@ -153,8 +175,49 @@ def add_e2e_tests(litani, test_dir, counter, output_dir, fast):
...
+            set_jobs_output = str(uuid.uuid4())
+            litani_add(
+                litani, counter,
+                command=collapse(f"""
+                    {e2e_test_dir / 'run'}
+                    --test-file {test_file}
+                    --litani {litani}
+                    --run-dir {run_dir}
+                    --operation set-jobs"""),
+                pipeline=f"End-to-end: {test_file.stem}",
```
ronakfof: Please update to `pipeline=f"e2e_{test_file.stem}",`

This is needed since `:` is considered an invalid character by the `upload-artifact` action in `run-tests` workflow.

test/run
karkhaz: `add_jobs_output` is the output of add-jobs, so everything that takes it as an input will run after `add-jobs`. However, if there are several of them, they won't be ordered with respect to each other. So I think what I wrote is correct?

doc/src/man/litani-set-jobs.scdoc
karkhaz: Then in the documentation let's suggest the easy thing. I think asking the user to use dashes is easier than underscore, first of all because of the shift key, and also because it's closer to the actual command-line name

nrdg42: If the user is getting the list of jobs from get-jobs, then they will be underscores already

test/run
nrdg42: Oh so youre suggesting that the other jobs still take add jobs as an input but the run-build one does as well? Yeah that would work, but diagram doesn't come out as clean.

karkhaz: Yes, understood. Agree that the redundant edges are a bit annoying... if there's a way to have the right order without them, I'd be happy for that, but otherwise I think the edges are harmless.

angelonakos (approved):

ronakfof (approved):

karkhaz (approved):
Awesome work. Looking forward to seeing this used 🙂

PR #132, opened 2022-03-14
Print out stderr when a job fails

### LINKED ISSUE #131: Emit error logs when a job fails
Hi,

It would be great if we could configure Litani to print the error logs emitted by any job that fails. This would really help debugging a CI run that fails without having to manually re-run all the tests locally. Local tests don't always behave like CI tests, for example, when there's a configuration issue or the failure is performance related, which makes it nearly impossible to debug without proper logs.

Thanks,
Celina

karkhaz:
The entire buffered stderr of a job that fails will now be printed to
the terminal after the failing command line. This is to help users
quickly debug these jobs without viewing the HTML report.

This commit fixes #131.

### DIFF
--- a/litani
+++ b/litani
@@ -772,6 +772,11 @@ async def exec_job(args):
             with litani.atomic_write(arg_file) as handle:
                 print(out_str, file=handle)
 
+    if out_data["wrapper_return_code"] and out_data["stderr"]:
+        print(
+            "\n".join([l.rstrip() for l in out_data["stderr"]]),
+            file=sys.stderr)
+
     timestamp("end_time", out_data)
     out_str = json.dumps(out_data, indent=2)
     logging.debug("run status: %s", out_str)


celinval:
You're fast! Thanks! Can you please also add an option to emit stdout?

angelonakos (approved):

ronakfof (approved):

karkhaz:
@celinval
> Can you please also add an option to emit stdout?

There's two ways this could be done, I'm guessing you probably want option 1?

1. A flag to `litani run-build` that makes *all* failing jobs dump both their stdout and stderr
2. A flag to `litani add-job` that makes that particular job dump its stdout if it fails

@markrtuttle do you also have an opinion here? I suppose I could even do both. But I'm not really sure if 2. is useful.

celinval:
Both would work for us. Option 1 is what we really want, but we could use option 2 to implement 1.

karkhaz:
Can I ask that we merge this and leave the stdout part for next release? It's a slightly weird reason: adding this option to either `run-build` or `add-jobs` would conflict with some work that @nrdg42 is doing, and Nick's work is pretty high priority and I'd rather not slow him down. I'd rather get this stderr patch and Nick's work merged and cut a release after that, and then do the stdout feature on top of Nick's work.

celinval:
For sure. We can redirect the stdout to stderr for now if we want to see everything. This is already huge! Thanks

PR #135, opened 2022-03-25
Upload report as artifact on test failure

ronakfof:
*Issue #, if available:*

*Description of changes:*

Litani Test Suite gets stuck sometimes on some specific proofs and
gets timed out. With this commit we can access those reports as
artifacts and dive into the issue instead of trying to recreate the
errors again locally.

[Example Run](https://github.com/awslabs/aws-build-accumulator/actions/runs/2041065375)

### DIFF
--- a/.github/workflows/run-tests.yaml
+++ b/.github/workflows/run-tests.yaml
@@ -16,5 +16,18 @@ jobs:
         run: |
           brew install ninja
           python3 -m pip install jinja2
+
       - name: Run Unit and e2e tests
         run: ./test/run
+        continue-on-error: true
+        timeout-minutes: 5
+
+      - name: Get absolute path to report dir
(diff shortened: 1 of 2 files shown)


karkhaz (approved):
Thanks. What breaks if the pipeline name contains whitespace?

ronakfof:
> Thanks. What breaks if the pipeline name contains whitespace?

Sorry i forgot to mention - 
Upload-artifacts path considers `:` as invalid character
```
Error: Artifact path is not valid: /artifacts/End-to-end: custom_stages/index.html.
Contains the following character:  Colon :
          
Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >,
Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n
```

nrdg42 (approved):

PR #137, opened 2022-03-30
Add concurrency-safe HTML report directory API

karkhaz:
This feature adds 3 new commands that allow Litani clients to access the HTML report, either for continuous viewing or modification.

- `litani print-html-dir` prints the path to a HTML report directory that will be continuously updated with new progress while `litani run-build` runs. This is the command for local users to use for viewing the HTML report in a browser, for example.
- `litani acquire-html-dir` prints the path to an exclusively-locked HTML report directory. After this command terminates, the printed-out directory will not be deleted or modified until `litani release-html-dir` is subsequently called. This is the command to use when you require prolonged exclusive access to HTML data, for example while doing a file transfer.
- `litani release-html-dir` releases the lock on a previously-acquired HTML report directory. This allows other processes to attempt to lock the directory, and also allows *litani run-build* to acquire the directory for garbage collection.

This feature gives users a safe way to get the path to a HTML report directory that meets their usage requirements, and without relying on internal Litani implementation details.

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License.
 
 
+import asyncio
 import dataclasses
 import datetime
 import enum
@@ -383,6 +384,32 @@ class ReportRenderer:
 
         litani.unlink_expired()
 
+
+
+async def release_html_dir(args):
+    lockable_dir = lib.litani.LockableDirectory(args.dir)
+    lockable_dir.release()
+
+
+async def acquire_html_dir(args):
+    remaining = args.timeout
+    while True:
+        html_dir = lib.litani.get_report_dir().resolve()
+        lockable_dir = lib.litani.LockableDirectory(html_dir)
+        if lockable_dir.acquire():
+            print(html_dir)
+            sys.exit(0)
+        remaining -= 1
+        if remaining == 0:
+            sys.exit(1)
+        await asyncio.sleep(1)
+
+
+async def print_html_dir(_):
+    html_dir = lib.litani.get_report_dir()
+    if not html_dir.exists():
+        sys.exit(1)
+    print(html_dir)
 
 
 def get_run(cache_dir):
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -583,14 +583,38 @@ def render(run, report_dir, pipeline_depgraph_renderer):
     with litani.atomic_write(temporary_report_dir / "index.html") as handle:
         print(page, file=handle)
 
+    # The sequence is:
+    #
+    # Initially:
+    # |\_ report_data_dir
+    # |    |\_ 1234-abcd [report data that has just been written]
+    # |    \__ 5678-olde
+    # \__ html  -symlink->  report_data_dir/5678-olde
+    #
+    # 1. 1234-abcd gets locked
+    #
+    # 2. report_data_dir
+    #    |- 1234-abcd
+    #    |- 5678-olde
+    #    html  -symlink->  report_data_dir/5678-olde
+    #    html.987-xzy  -symlink->  1234-abcd
+    #
+    # 3. report_data_dir
+    #    |- 1234-abcd
+    #    |- 5678-olde
+    #    html  -symlink->  1234-abcd
+    #
+    # 4. 1234-abcd gets unlocked
+    #
+    # 5. 5678-olde gets marked as expired.
+
+    locked_dir = litani.LockableDirectory(temporary_report_dir)
 
     temp_symlink_dir = report_dir.with_name(report_dir.name + str(uuid.uuid4()))
     os.symlink(temporary_report_dir, temp_symlink_dir)
     os.rename(temp_symlink_dir, report_dir)
 
-    # Release lock so that other processes can read from this directory
-    new_report_dir = litani.LockableDirectory(report_dir.resolve())
-    new_report_dir.release()
+    locked_dir.release()
 
     if old_report_dir_path.exists():
         old_report_dir.expire()
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -297,6 +297,109 @@ class StatsGroupRenderer:
 
 
 
+def render(run, report_dir, pipeline_depgraph_renderer):
+    temporary_report_dir = litani.get_report_data_dir() / str(uuid.uuid4())
+    temporary_report_dir.mkdir(parents=True)
+    old_report_dir_path = litani.get_report_dir().resolve()
+    old_report_dir = litani.ExpireableDirectory(old_report_dir_path)
+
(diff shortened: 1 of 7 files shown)


karkhaz:
The only really important commit is the last one, "Add HTML report directory API".

markrtuttle (approved):

ronakfof:
Running ./test/run` gives error
```
ERROR:root:Could not find a pointer to a litani cache. Did you forget to run `litani init`?
```
both in CI and locally.

---
Running `./litani init --project-name test` locally fails as well with 
```
ERROR:root:Could not find a pointer to a litani cache. Did you forget to run `litani init`?
```

lib/litani_report.py
```diff
@@ -410,6 +411,35 @@ def __call__(self, run):
...
+    lockable_dir.release()
+
+
+async def acquire_html_dir(args):
+    remaining = args.timeout
+    while True:
+        html_dir = lib.litani.get_report_dir().resolve()
+        lockable_dir = lib.litani.LockableDirectory(html_dir)
+        try:
+            if lockable_dir.acquire():
```
angelonakos: Going on a slight tangent.

I was looking at `lib/litani.py`:

```python
class LockableDirectory:
    """POSIX-compliant directory locking"""

# ...

    def acquire(self):
        """Return True if acquisition was successful, False otherwise"""
        try:
            self._lock_file.unlink()
        except OSError:
            return False
        else:
            return True
```

Why does `acquire()` handle only the case of `OSError` and not `FileNotFoundError` ?

karkhaz: `OSError` is a superclass of `FileNotFoundError`, which means that the catch here is redundant. Thanks.

litani
```diff
@@ -288,6 +288,37 @@ def get_args():
...
+            "metavar": "D",
+    }]:
+        flags = arg.pop("flags")
+        release_html_dir_pars.add_argument(*flags, **arg)
+
+    acquire_html_dir_pars = subs.add_parser("acquire-html-dir",
+        help="Print the path to a locked HTML report directory")
+    acquire_html_dir_pars.set_defaults(func=lib.litani_report.acquire_html_dir)
+    for arg in [{
+            "flags": ["-t", "--timeout"],
```
angelonakos, line 229: This flag:
1. could become required.
2. could remove its default of `0`, so that Litani clients don't issue a command that runs forever.

lib/litani_report.py
```diff
@@ -297,6 +298,148 @@ def render(self, job_filter, template_name):
...
+                print(html_dir)
+                sys.exit(0)
+        except FileNotFoundError:
+            pass
+        remaining -= 1
+        if remaining == 0:
+            sys.exit(1)
+        await asyncio.sleep(1)
```
angelonakos, line 405: Why is the block of 1 second via `await asyncio.sleep(1)` being added?

karkhaz: We want to retry every second, there's no need to retry as fast as possible?

`time.sleep` isn't the right call here because we're inside an `async` method. What `asyncio.sleep` does is to put the current green thread to sleep, but allow other green threads to proceed if they have work to do. Now in this particular case, there are actually no other green threads (the only litani command that is actually multi-threaded is `run-build`, since it both needs to run the command and do a bunch of monitoring like parallelism and so on). But it still looks very suspicious to have `time.sleep` in an async method, so I'd rather not have it

litani
karkhaz: Done, thanks.

litani
```diff
@@ -672,17 +704,18 @@ async def run_build(args):
...
     run = litani_report.get_run_data(cache_dir)
     lib.validation.validate_run(run)
-    pipeline_degraph_renderer = litani_report.PipelineDepgraphRenderer(
+    report_dir = lib.litani.get_report_dir()
+    pipeline_depgraph_renderer = litani_report.PipelineDepgraphRenderer(
         should_render=not args.no_pipeline_dep_graph)
-    litani_report.render(run, report_dir, pipeline_degraph_renderer)
+    render = litani_report.ReportRenderer(
+        report_dir, pipeline_depgraph_renderer)
+    render(run)
```
nrdg42, line 648: I'm not used to seeing this type of syntax being used where an instance of a class is itself a callable object. I know python has support for the `__call__` method as you've utilized here, but is this a common pattern? It just reads to me as though you're convoluting the two

lib/litani_report.py
```diff
@@ -297,6 +298,145 @@ def render(self, job_filter, template_name):
 
 
 
+@dataclasses.dataclass
+class ReportRenderer:
+    report_dir: pathlib.Path
+    pipeline_depgraph_renderer: PipelineDepgraphRenderer
+
+
+    def __call__(self, run):
```
nrdg42, line 307: This is a long and complicated function, is there anything we can abstract into smaller chunks to make it easier to read and maintain?

lib/litani_report.py
```diff
@@ -297,6 +298,145 @@ def render(self, job_filter, template_name):
...
+        #    html.987-xzy  -symlink->  1234-abcd
+        #
+        # 3. report_data_dir
+        #    |- 1234-abcd
+        #    |- 5678-olde
+        #    html  -symlink->  1234-abcd
+        #
+        # 4. 1234-abcd gets unlocked
+        #
+        # 5. 5678-olde gets marked as expired.
```
nrdg42: I didn't really find this comment very helpful compared to just reading the code. I think one line comments above the code to tell what you are doing would be better. I don't think I got this 100% accurate so adjust these comments how you'd see fit.
```python
        # Lock the directory where we've just rendered the newest report to
        locked_dir = litani.LockableDirectory(temporary_report_dir)

        # Create a temporary symlink to the new report dir
        temp_symlink_dir = self.report_dir.with_name(
            self.report_dir.name + str(uuid.uuid4()))
        os.symlink(temporary_report_dir, temp_symlink_dir)

        # Replace the report_dir symlink with new report
        os.rename(temp_symlink_dir, self.report_dir)

        # Release lock on new report directory
        locked_dir.release()

        # Expire and unlink old report directory
        try:
            old_report_dir.expire()
        except FileNotFoundError:
            pass

        litani.unlink_expired()
```

ronakfof (approved):

litani
karkhaz: I wouldn't say it's common as such... but I like it better than the alternative, which is to have useless method names like
```
renderer.render()
renderer.do()
```
etc. If the whole point of an object is to *do* something, rather than to store data, then I think instances should be named as verbs and have a `__call__` method.

lib/litani_report.py
karkhaz: Yes, but I'd rather not do that in this PR, this PR just moves it for refactoring... and ultimately I'm planning for this implementation to go away reasonably soon.

PR #141, opened 2022-04-06
Use more reliable timeout for HTML acquire

karkhaz:

### DIFF
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -392,15 +392,17 @@ async def release_html_dir(args):
 
 
 async def acquire_html_dir(args):
-    remaining = args.timeout
+    if args.timeout:
+        end = datetime.datetime.now() + datetime.timedelta(seconds=args.timeout)
+    else:
+        end = datetime.datetime.max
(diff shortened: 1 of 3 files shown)


angelonakos (approved):

lib/litani_report.py
```diff
@@ -392,15 +392,16 @@ async def release_html_dir(args):
 
 
 async def acquire_html_dir(args):
-    remaining = args.timeout
+    start = datetime.datetime.now()
+    timeout = args.timeout or 10 ** 10    # >300 years
+    end = start + datetime.timedelta(seconds=timeout)
```
nrdg42: 300 years is probably a sensible default, but why not
```python
end = datetime.datetime.max
if args.timeout:
    end = start + datetime.timedelta(seconds=args.timeout)
```
Just feels like the cleaner way of doing it

karkhaz: I didn't know about `datetime.datetime.max`, thank you!

nrdg42 (approved):

PR #152, opened 2022-04-12
Do not create HTML docs of Litani man pages if a flag is specified

angelonakos:
*Issue #, if available:*

*Description of changes:*

Running `doc/configure` with `--no-html` allows one
to ultimately build only the man equivalent of the
Litani documentation.

### DIFF
--- a/doc/configure
+++ b/doc/configure
@@ -14,8 +14,9 @@
 # permissions and limitations under the License.
 
 
-import sys
+import argparse
 import pathlib
+import sys
 
 
 DOC_DIR = pathlib.Path(sys.path[0]).resolve()
@@ -89,6 +90,19 @@ RULES = [{
 }]
 
 
+def get_args():
+    pars = argparse.ArgumentParser(description="Build documentation for Litani")
+    for arg in [{
+            "flags": ["--no-html"],
+            "action": "store_false",
+            "help": "Do not build a version of the documentation in HTML",
+            "dest": "gen_html",
+    }]:
+        flags = arg.pop("flags")
+        pars.add_argument(*flags, **arg)
+    return pars.parse_args()
+
+
 def make_html_unique(man_name, html_man, html_mans, builds):
     html_unique = HTML_UNIQUE_DIR / f"{man_name}.html"
     builds.append({
@@ -118,7 +132,7 @@ def man_to_html(man_name, man_out, builds):
 
 
 def convert_man_dir_to_man(
-        src_dir, dst_dir, rule, html_mans, builds, extra_inputs=None):
+        src_dir, dst_dir, rule, html_mans, builds, gen_html, extra_inputs=None):
     for man in (src_dir).iterdir():
         man_name = str(man.name).split(".", 1)[0]
         if man.suffix == ".scdoc":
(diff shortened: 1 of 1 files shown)


doc/configure
```diff
@@ -118,7 +131,8 @@ def man_to_html(man_name, man_out, builds):
 
 
 def convert_man_dir_to_man(
-        src_dir, dst_dir, rule, html_mans, builds, extra_inputs=None):
+        src_dir, dst_dir, rule, html_mans, builds, extra_inputs=None,
+        html=True):
```
karkhaz: No need for a default argument, since an argument is always being passed

doc/configure
```diff
@@ -171,34 +186,46 @@ def add_litani_7(builds, html_mans):
...
 
 
 def main():
+    args = get_args()
     builds = []
     html_mans = []
 
     convert_man_dir_to_man(
-        SRC_DIR / "man", MAN_DIR, "sc_to_man", html_mans, builds)
+        SRC_DIR / "man",
```
karkhaz: No need for one argument on each line.

karkhaz (changes requested):

doc/configure
angelonakos: I switched to having as many arguments per line and wrapping at the limit. Is this preferable?

karkhaz (approved):

doc/configure
```diff
@@ -118,7 +131,7 @@ def man_to_html(man_name, man_out, builds):
 
 
 def convert_man_dir_to_man(
-        src_dir, dst_dir, rule, html_mans, builds, extra_inputs=None):
+        src_dir, dst_dir, rule, html_mans, builds, html, extra_inputs=None):
```
nrdg42: `html` is not a very good name for a boolean value

angelonakos: What would you go with?

karkhaz: `generate_html` or `gen_html`?

angelonakos: Thanks Kareem. I went with `gen_html`.

doc/configure
```diff
@@ -171,34 +185,39 @@ def add_litani_7(builds, html_mans):
...
-            TEMPLATE_DIR / "index.jinja.html",
-        ],
-        "outputs": [DOC_DIR / "out" / "html"/ "index.html"],
-        "rule": "build_html_doc",
-        "variables": {
-            "html-mans": " ".join([str(h) for h in html_mans]),
-            "man-html-dir": HTML_MAN_SRC_DIR,
-        }
-    })
+    if html_mans:
```
karkhaz: I think it's clearer to use `if not args.no_html`.

`if html_mans` is slightly preferable for the following reason: if (due to a programming error) that array was non-empty even though `--no-html` is passed, then ninja will refuse to run because this job's dependencies don't exist. So it helps us to catch mistakes in this script.

However, I think it's more important for this to be readable.

angelonakos: Right. I do agree that it's more clear to first-time reader of `main()`

doc/configure
```diff
@@ -89,6 +90,18 @@ RULES = [{
 }]
 
 
+def get_args():
+    pars = argparse.ArgumentParser(description="Build documentation for Litani")
+    for arg in [{
+            "flags": ["--no-html"],
+            "action": "store_true",
+            "help": "Do not build a version of the documentation in HTML",
```
nrdg42: Since you use `not` everywhere you use `args.no_html` below, maybe this is a cleaner implementation:

```python
    for arg in [{
            "flags": ["--no-html"],
            "action": "store_false",
            "help": "Do not build a version of the documentation in HTML",
            "dest": "gen_html"
    }]:
```
That way you can just jump directly to `args.gen_html` without having to use `not args.no_html` first

nrdg42 (approved):

PR #157, opened 2022-05-04
Fix litani man pages

ronakfof:
*Issue #, if available:*

*Description of changes:*
Read comments for explanation on changes

### DIFF
--- a/doc/configure
+++ b/doc/configure
@@ -134,7 +134,7 @@ def man_to_html(man_name, man_out, builds):
 def convert_man_dir_to_man(
         src_dir, dst_dir, rule, html_mans, builds, gen_html, extra_inputs=None):
     for man in (src_dir).iterdir():
-        man_name = str(man.name).split(".", 1)[0]
+        man_name = man.stem
         if man.suffix == ".scdoc":
             with open(man) as fp:
                 line = fp.readline().rstrip()
@@ -207,7 +207,7 @@ def main():
     add_litani_7(builds, html_mans, args.gen_html)
 
     for h in html_mans:
-        if "litani.7" in str(h.resolve()):
+        if "litani.7" in h.name:
             html_mans.insert(0, html_mans.pop(html_mans.index(h)))
 
     if args.gen_html:
--- a/doc/configure
+++ b/doc/configure
@@ -206,6 +206,10 @@ def main():
 
     add_litani_7(builds, html_mans, args.gen_html)
 
+    for h in html_mans:
+        if "litani.7" in str(h.resolve()):
+            html_mans.insert(0, html_mans.pop(html_mans.index(h)))
+
     if args.gen_html:
         builds.append({
             "inputs": html_mans + [


angelonakos (approved):

doc/configure
```diff
@@ -134,7 +134,7 @@ def man_to_html(man_name, man_out, builds):
 def convert_man_dir_to_man(
         src_dir, dst_dir, rule, html_mans, builds, gen_html, extra_inputs=None):
     for man in (src_dir).iterdir():
-        man_name = str(man.name).split(".", 1)[0]
```
ronakfof, line 137: This was required as the logic `str(man.name).split(".", 1)[0]` was returning -

`litani-outcome-table` INSTEAD OF `litani-outcome-table.json`
AND
`litani-run` INSTEAD OF `litani-run.json`
```

File: doc/configure [hunk 2]
```diff
@@ -206,6 +206,10 @@ def main():
 
     add_litani_7(builds, html_mans, args.gen_html)
 
+    for h in html_mans:
+        if "litani.7" in str(h.resolve()):
+            html_mans.insert(0, html_mans.pop(html_mans.index(h)))
+
```
ronakfof, line 212: This change is made in order to bring `litani.7` at top of `html_mans` list so that it can be parsed first and stay on top in report's navigation.

doc/configure
nrdg42: Is this not the desired behavior? It seems like `man.stem` does the same thing as `str(man.name).split(".", 1)[0]`

doc/configure
nrdg42: Could you say `if h.name == "litani.7"`? Doing a string comparison would be more efficient than doing a substring search

doc/configure
ronakfof: We need `litani-outcome-table.json` and `litani-run.json` as output but `str(man.name).split(".", 1)[0]` returns `litani-outcome-table` and `litani-run` because of which the schema for both in documentation was breaking earlier.

doc/configure
ronakfof: Can't do a string comparison as `h` here is a `pathlib.Path` which contains `litani.7`.

doc/configure
nrdg42: How is what `man.stem` returns any different is my question

doc/configure
nrdg42: So what you're saying is that `litani.7` is not the name of a page? Is it a directory or something?

doc/configure
ronakfof: For a file named `litani-outcome-table.json.scdoc` -

- [x] `man.stem` returns `litani-outcome-table.json`

- [ ] `str(man.name).split(".", 1)[0]` returns `litani-outcome-table`

For a file named `litani-run.json.scdoc` -

- [x] `man.stem` returns `litani-run.json`

- [ ] `str(man.name).split(".", 1)[0]` returns `litani-run`

doc/configure
ronakfof: value of `h` that we are looking for is `PosixPath('/Users/ronakf/Desktop/litani/doc/tmp/html_unique/litani.7.html')`
I check if `litani.7` is present and then bring it to top of list `html_mans` so that it is rendered first in template.

doc/configure
nrdg42: Thank you, I did not have the context that they were .scdoc files

doc/configure
nrdg42: Would it then make sense to say at least `if "litani.7" in h.name`? That way you don't need to search through the entire path from the root to the file

nrdg42: Or why don't you say even `if "litani.7.html" == h.name?`

ronakfof: Makes sense. Fixed.

nrdg42 (approved):

PR #159, opened 2022-05-11
Cache path to cache directory

ronakfof:
This reverts commit 90eb73555aaa584f59cc635bfc308773491be0b3.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/lib/litani.py
+++ b/lib/litani.py
@@ -14,7 +14,6 @@
 
 import asyncio
 import contextlib
-import dataclasses
 import json
 import logging
 import os
@@ -68,7 +67,6 @@ class TimeoutExpired(Exception):
     pass
 
 
-@dataclasses.dataclass
 class CacheDir:
     cache_dir_path: pathlib.Path = None
 
@@ -76,14 +74,8 @@ class CacheDir:
     def get(lookup_path=os.getcwd()):
         if CacheDir.cache_dir_path:
             return CacheDir.cache_dir_path
-        def cache_pointer_dirs():
-            current = pathlib.Path(lookup_path).resolve(strict=True)
-            yield current
-            while current.parent != current:
-                current = current.parent
-                yield current
 
-        for possible_dir in cache_pointer_dirs():
+        for possible_dir in CacheDir.cache_pointer_dirs(lookup_path):
             logging.debug(
                 "Searching for cache pointer in %s", possible_dir)
             possible_pointer = possible_dir / CACHE_POINTER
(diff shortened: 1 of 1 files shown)


angelonakos (approved):

angelonakos:
This PR is bringing back changes that were introduce in #119

nrdg42:
Please update commit message with:

* Reasons for this change (what it is doing and why)
* Why it was reverted in the first place
* Why it is okay to include now

And I would personally prefer it if the commit message was not "Revert Revert", just rename it "Cache path to cache directory"

ronakfof:
> Reasons for this change (what it is doing and why)
> And I would personally prefer it if the commit message was not "Revert Revert", just rename it "Cache path to cache directory"

Updated commit message.

lib/litani.py
```diff
@@ -67,6 +68,48 @@ class TimeoutExpired(Exception):
     pass
 
 
+@dataclasses.dataclass
```
karkhaz: I'm not sure that this should be a dataclass. This would synthesize an `__init__` method, when in fact we're not intending to ever instantiate this class. I think `cache_dir_path` should be treated as a static variable, which you're already doing... so I think dataclass is unnecessary

lib/litani.py
```diff
@@ -67,6 +68,48 @@ class TimeoutExpired(Exception):
...
+class CacheDir:
+    cache_dir_path: pathlib.Path = None
+
+    @staticmethod
+    def get(lookup_path=os.getcwd()):
+        if CacheDir.cache_dir_path:
+            return CacheDir.cache_dir_path
+        def cache_pointer_dirs():
```
karkhaz: This can be a separate static method

lib/litani.py
```diff
@@ -67,6 +68,48 @@ class TimeoutExpired(Exception):
...
+            while current.parent != current:
+                current = current.parent
+                yield current
+
+        for possible_dir in cache_pointer_dirs():
+            logging.debug(
+                "Searching for cache pointer in %s", possible_dir)
+            possible_pointer = possible_dir / CACHE_POINTER
+            try:
+                if possible_pointer.exists():
```
karkhaz, line 83: In a separate commit, could you refactor this code so that it's not so deeply nested?

```
if not possible_pointer.exists():
    continue

# de-indent the rest of the method
```

karkhaz (changes requested):
Fantastic, thank you! This is much more sensible.

karkhaz:
Wait, I thought this was familiar. I agree with Nick, this needs an explaination... I wish I wrote down why it was reverted, that's really annoying. Does anyone remember? There must have been a reason if we did it just before the release.

angelonakos (approved):

nrdg42 (approved):

PR #164, opened 2022-05-19
Refactor Litani commands

nrdg42:
Refactor the base litani file. Get subparsers from the lib folder and
move entry points into files in the lib folder

*Issue #, if available:*

*Description of changes:*


litani
```diff
@@ -95,220 +63,19 @@ def get_args():
...
+    lib.jobs.add_add_jobs_subparser(subs)
+    lib.jobs.add_get_jobs_subparser(subs)
+    lib.jobs.add_set_jobs_subparser(subs)
+    lib.jobs.add_transform_jobs_subparser(subs)
+    lib.graph.add_subparser(subs)
+    lib.run_build.add_subparser(subs)
+    lib.exec.add_subparser(subs)
+    lib.capabilities.add_subparser(subs)
+    lib.run_printer.add_subparser(subs)
+    lib.litani_report.add_print_html_dir_subparser(subs)
```
karkhaz: Suggest:
* the entry points are all in a directory called `entry_points`
* the code here iterates over all the modules and calls `add_subparser`

karkhaz (changes requested):

litani
nrdg42: From our offline conversation: Its not clear what the `entry_points` directory would look like. It could have a file for each individual subcommand, some with multiple subcommands, or a single big file, which may kind of defeat the purpose of the refactoring. What I've done instead was to iterate over the `lib` directory and call `add_subparser` on each file which contains that method

karkhaz (approved):
Thanks so much for this, long overdue :)

angelonakos (approved):

PR #167, opened 2022-05-24
Summarize build results by printing a JSON object

angelonakos:
*Issue #, if available:*

*Description of changes:*

A flag was added to `run-build` that summarizes the results of a build by printing a JSON object.
    
A user will be able to get a breakdown of statuses across pipelines as well as a breakdown of the
status of each pipeline.

### DIFF
--- a/lib/post_build.py
+++ b/lib/post_build.py
@@ -1,0 +1,34 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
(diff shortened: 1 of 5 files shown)


lib/post_build.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_summary(run_dict):
+    # The list of possible statuses can be viewed in the man page for run.json
+    status_summary_map = {}
+    pipeline_status_map = {}
+    for pipeline in run_dict["pipelines"]:
+        status_name = pipeline["status"]
+        status_name_pretty = status_name.title().replace("_", " ")
+        if status_name_pretty not in status_summary_map:
```
karkhaz:
```
try:
  status_summary_map[status_name_pretty] += 1
except KeyError:
    status_summary_map[status_name_pretty] = 1

karkhaz (changes requested):
I don't think this application-level reporting belongs here. Different users of Litani would want different kinds of reports, and I wouldn't want to bake this particular JSON report into the CL interface... once it's in, we can't change it, even if somebody wants something slightly different.

I think applications should just run `dump-run` or `get-jobs` (preferably the latter) and use that for whatever output they want.

lib/post_build.py
angelonakos: Done

angelonakos:
Canceling

PR #169, opened 2022-06-11
Refactor test suite & improve start up time

karkhaz:
The test suite now runs `litani add-job` asynchronously. On a 12-core
machine, this decreases test startup time by 22x.

### DIFF
--- a/test/run
+++ b/test/run
@@ -15,8 +15,11 @@
 
 
 import argparse
+import asyncio
+import dataclasses
 import importlib
 import logging
+import math
 import os
 import pathlib
 import re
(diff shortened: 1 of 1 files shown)


test/run
```diff
@@ -44,10 +47,11 @@ def get_args():
     return pars.parse_args()
 
 
-def run_cmd(cmd):
-    try:
-        subprocess.run([str(c) for c in cmd], check=True)
-    except subprocess.CalledProcessError:
+async def run_cmd(cmd):
+    proc = await asyncio.create_subprocess_exec(
+        str(cmd[0]), *[str(arg) for arg in cmd[1:]])
```
angelonakos, line 52: Would `proc = await asyncio.create_subprocess_exec(*[str(arg) for arg in cmd])` also work?

angelonakos (approved):

test/run
karkhaz: No, the command and arguments are two separate arguments to the [`create_subprocess_exec`](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.create_subprocess_exec) call.

PR #171, opened 2022-06-27
Fix incorrect RC with --fail-on-pipeline-failure

karkhaz:
This fixes an important bug in --fail-on-pipeline-failure. Prior to this
commit, `litani run-build` would only terminate abnormally with this
flag set when ninja terminated abnormally. This commit fixes this
behavior so that `litani run-build` instead checks whether any pipeline
failed, and terminates abnormally if so.

### DIFF
--- a/lib/run_build.py
+++ b/lib/run_build.py
@@ -222,7 +222,7 @@ async def run_build(args):
     if args.fail_on_pipeline_failure:
         for _ in itertools.filterfalse(
                 lambda pipe: pipe["status"] != "fail", run["pipelines"]):
-            sys.exit(1)
+            sys.exit(10)
         sys.exit(0)
 
     sys.exit(0)
--- a/lib/run_build.py
+++ b/lib/run_build.py
@@ -12,6 +12,7 @@
 # permissions and limitations under the License
 
 import datetime
+import itertools
 import json
 import logging
 import os
@@ -219,6 +220,9 @@ async def run_build(args):
               f"file://{run_info['latest_symlink']}/html/index.html")
 
     if args.fail_on_pipeline_failure:
-        sys.exit(0 if runner.was_successful() else 1)
+        for _ in itertools.filterfalse(
+                lambda pipe: pipe["status"] != "fail", run["pipelines"]):
+            sys.exit(1)
+        sys.exit(0)
 
     sys.exit(0)
--- a/test/e2e/run
+++ b/test/e2e/run
@@ -29,17 +29,18 @@ DESCRIPTION = "Execute a single Litani run and test the resulting run.json"
 EPILOG = "See test/e2e/README for the organization of this test suite"
 
 
-def run_cmd(cmd):
+def run_cmd(cmd, check):
     cmd_list = [str(c) for c in cmd]
     print(" ".join(cmd_list))
     try:
-        subprocess.run(cmd_list, check=True)
+        proc = subprocess.run(cmd_list, check=check)
     except subprocess.CalledProcessError:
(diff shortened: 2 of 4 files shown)


lib/run_build.py
```diff
@@ -219,6 +220,9 @@ async def run_build(args):
               f"file://{run_info['latest_symlink']}/html/index.html")
 
     if args.fail_on_pipeline_failure:
-        sys.exit(0 if runner.was_successful() else 1)
+        for _ in itertools.filterfalse(
```
markrtuttle, line 223: Prefer 
```
if any(pipe["status"] == "fail" for pipe in run["pipelines"]):
  sys.exit(1)
```

test/e2e/run
```diff
@@ -29,17 +29,18 @@ DESCRIPTION = "Execute a single Litani run and test the resulting run.json"
 EPILOG = "See test/e2e/README for the organization of this test suite"
 
 
-def run_cmd(cmd):
+def run_cmd(cmd, check):
```
markrtuttle, line 32: Should check have a default value?  In other functions the default value is check=True but the only invocation of these functions is with top-level invocation check=False.  Suggests False should be the default value everywhere.

test/e2e/run
```diff
@@ -130,8 +131,27 @@ def set_jobs(litani, run_dir, mod):
...
     args = mod.get_run_build_args()
-    run_litani(
-        litani, "run-build", *args.get("args", []), **args.get("kwargs", {}))
+    proc = run_litani(
+        litani, "run-build",
+        *args.get("args", []), check=False, **args.get("kwargs", {}))
+
+    try:
+        expected_rc = mod.get_run_build_return_code()
+        if any((
```
markrtuttle: This does not test that expected_rc == proc.returncode as your commit messages says it will.

markrtuttle: Does this address the customer problem motivating this change?  I thought the customer concern was proof failure did not cause the run script to fail in the customer context (while it should not cause the run script to fail in the standard continuous integration context).

markrtuttle (changes requested):

lib/run_build.py
karkhaz: No need for `any`, even just the list would suffice (and would be faster) and I originally did that:
```
if [pipe["status"] == "fail" for pipe in run["pipelines"]]:
```
However, I prefer my version because it breaks out of the loop upon finding the first failure, rather than iterating along the entire list of pipelines every time. I'm trying really hard not to introduce needless slowdowns into this codebase, it's already bad enough.

test/e2e/run
karkhaz: This function `run_cmd` is only called from one place---`run_litani` (line 64 below). Since that call gives an explicit argument to `check`, I don't see a need for this parameter to be optional. If I start calling this function from other places and the argument becomes cumbersome, I'd remove it at that point

test/e2e/run
karkhaz: Done, thank you.

Yes, the run script wasn't failing because Litani wasn't, the run script was just checking Litani's return code rather than reading run.json. This fixes that bug.

ronakfof (approved):

markrtuttle (approved):
I am approving this code, but there is a lingering issue that needs to be addressed in another code change:  This code does nothing to distinguish between proof failures and any other failure, it simply returns an error code 10 if any pipeline fails, doesn't matter if it was a goto-cc command or a cbmc command.  I'm under the impression that the motivation for this pull request is to fail if a proof fails.

PR #173, opened 2022-07-11
Update redirect behavior for stages on dashboard page

angelonakos:
*Issue #, if available:*

*Description of changes:*

If a pipeline lacks jobs for a certain stage, then
the corresponding "bar" displayed on the dashboard
page should not be "clickable" and it should not
redirect to an "artifacts page" since no artifacts
will exist.

### DIFF
--- a/templates/dashboard.jinja.html
+++ b/templates/dashboard.jinja.html
@@ -446,13 +446,15 @@ p {
         {% for stage in pipe["ci_stages"] %}
         <div class="pipeline-stage">
 
-          <a class="stage-artifacts-link {{ stage['status'] }}"
-             href="{{ stage['url'] }}/index.html">
-            <div class="stage-bar">
-              <div class="progress" style="width: {{ stage['progress'] }}%">
-              </div><!-- class="progress" -->
-            </div><!-- class="stage-bar" -->
-          </a>
+          {% if stage['jobs'] %}
+            <a class="stage-artifacts-link {{ stage['status'] }}"
+              href="{{ stage['url'] }}/index.html">
+              <div class="stage-bar">
+                <div class="progress" style="width: {{ stage['progress'] }}%">
+                </div><!-- class="progress" -->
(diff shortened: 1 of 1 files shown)


templates/dashboard.jinja.html
```diff
@@ -446,13 +446,20 @@ <h1>
         {% for stage in pipe["ci_stages"] %}
         <div class="pipeline-stage">
 
-          <a class="stage-artifacts-link {{ stage['status'] }}"
-             href="{{ stage['url'] }}/index.html">
+          {% if stage['jobs'] %}
```
karkhaz, line 449: Just wrap the `<a>` and `</a>` in `{% if %}`, no need for `elif`. This implementation duplicates the `stage-bar` div

karkhaz (changes requested):

templates/dashboard.jinja.html
angelonakos: Done.

I would like to point out a visual difference between these 2 approaches.

If we just wrap the `<a>` and `</a>` in a `{% if %}` then the dashboard page looks like:

image


If we use an `elif` then the dashboard page looks like:

image


The "empty bars" in the second image suggest that no jobs have been added to those pipelines' `build` and `test` stages.

karkhaz: That doesn't seem right... I just tried it now, it looks fine:
Screenshot 2022-07-12 at 19 10 14

I think you deleted the `<a>` tag but also everything inside it?

As you see in the inspector, I just deleted the `<a>` that is inside the `pipeline-stage` and that surrounds the `stage-bar`, but I kept `stage-bar`, that's what I'm proposing to do here. So you'd wrap the `<a>` and `</a>` in `{% if %}` but keep the stuff that is currently inside `<a>`... sorry that I wasn't clear the first time.

karkhaz (changes requested):

PR #177, opened 2022-08-02
Modify behavior of acquire-html-dir command

angelonakos:
*Issue #, if available:*

*Description of changes:*

If a build has finished, then a lock will not be acquired (and consequently the path to the HTML directory will not be printed to stdout) until the report has finished rendering.

### DIFF
--- a/lib/litani.py
+++ b/lib/litani.py
@@ -58,6 +58,22 @@ class ExpireableDirectory:
 
 
 
+class ReportRendering:
+    """This class is to mark the completion of the report rendering process"""
+
+    def __init__(self):
+        self._touchfile = get_cache_dir().resolve() / ".litani-completed"
+
+
+    def complete(self):
+        self._touchfile.touch()
+
+
+    def has_completed(self):
+        return self._touchfile.exists()
+
+
+
 class AcquisitionFailed(Exception):
     pass
 
--- a/lib/litani_report.py
+++ b/lib/litani_report.py
@@ -448,16 +448,22 @@ async def acquire_html_dir(args):
         end = datetime.datetime.now() + datetime.timedelta(seconds=args.timeout)
     else:
         end = datetime.datetime.max
-    while True:
+    while datetime.datetime.now() <= end:
         html_dir = lib.litani.get_report_dir().resolve()
         lockable_dir = lib.litani.LockableDirectory(html_dir)
-        if lockable_dir.acquire():
-            print(html_dir)
-            sys.exit(0)
-        if datetime.datetime.now() > end:
-            sys.exit(1)
-        await asyncio.sleep(1)
-
+        if not lockable_dir.acquire():
+            await asyncio.sleep(1)
+            continue
+        with open(html_dir / "run.json") as handle:
+            run = json.load(handle)
+            is_completed_run = run["status"] != "in_progress"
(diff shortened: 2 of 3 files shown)


lib/litani_report.py
```diff
@@ -448,17 +448,34 @@ async def acquire_html_dir(args):
...
+        if run["status"] != "in_progress":
+            cannot_acquire = True
+            while datetime.datetime.now() < end:
+                if (html_dir / "index.html").exists():
+                    cannot_acquire = False
+                    break
+                await asyncio.sleep(1)
+            if cannot_acquire:
+                sys.exit(1)
+            lock(html_dir)
```
ronakfof: why lock here?

karkhaz (changes requested):
This almost works. Thanks!

It's still racy. We have to take the lock *before* attempting to read run.json or index.html, because otherwise Litani is free to garbage-collect the directory while we're trying to read it.

But the general idea works. I think this is the correct algorithm:
```
lockable_dir = lib.litani.LockableDirectory(html_dir)
while not timed out yet:
    if not lockable_dir.acquire():
        sleep(1)
        continue
    if run is completed and final report is not yet written:
        lockable_dir.release()
        sleep(1)
        continue
    print(html_dir)
    sys.exit(0)
sys.exit(1)
```
however, there's one more problem. You're currently assuming that just because `index.html` has been written, that means that the final report has been written. But in fact that index.html might have been come from a rendering call that happened before the run finished.

One possible solution to this would incorporate your touchfile idea from the PR that you closed. The difference with your PR is that the touchfile isn't being used as public API. The way this would work is: run build would terminate and join the render thread; then render the report one last time (that's the final report), and once that's complete, write a touchfile (in the cache dir not the report dir). Meanwhile, this code would check for the existance  of the touchfile, but otherwise use the same logic as what you wrote, i.e. only print the report path if either the run is incomplete, or if it's complete and the final report has been written.

lib/litani_report.py
```diff
@@ -448,16 +448,23 @@ async def acquire_html_dir(args):
...
-            sys.exit(1)
-        await asyncio.sleep(1)
-
+        if not lockable_dir.acquire():
+            await asyncio.sleep(1)
+            continue
+        with open(html_dir / "run.json") as handle:
+            run = json.load(handle)
+            is_completed_run = run["status"] != "in_progress"
+        has_rendered = (html_dir.parent.parent / ".litani-completed").exists()
```
karkhaz: This should use `is_completed()`

lib/litani.py
```diff
@@ -58,6 +58,22 @@ def is_expired(self):
 
 
 
+class ReportRenderingCompletion:
+    """This class is to mark the completion of the report rendering process"""
+
+    def __init__(self, path: pathlib.Path):
```
karkhaz: I don't think this class should take a `path` argument. The argument makes it easier to write unit tests because you can then pass a mock object, but this class is so simple that it doesn't need tests. At the same time, the only `path` that ever gets passed is the `cache_dir`, so I suggest that you call `get_cache_dir` from `__init__`. That will simplify both call sites.

karkhaz (changes requested):
This looks like it should work. Thanks!

karkhaz (approved):

ronakfof (approved):

